PR #11, opened 2017-07-07
Glossary updates, new test directory with some examples

nealmcb:
Is /test/ a good place to put test data for now, as referred to from our glossary, or is there a place that's more standard for F&F? Actually thinking about it a bit, /data/ would be better, with code in /test/, but again such things vary widely in different coding communities.


ranweiler:
I'd suggest putting various bits of example data _used for documentation_ in a subfolder with a name like `docs/example` or `docs/sample`. Actual software subprojects can manage their own pieces of sample data which are used for testing.

nealmcb:
I agree, though I imagine that we would have actual tests that work off at least the CVR example, so it isn't just for documentation.

ranweiler:
You're right, we will have tests like that. I propose thinking of documentation-level examples as totally separate. The individual (code) subprojects can add test data as needed, which they will "own". If we find we are duplicating data, then we can put it in a shared top-level directory when we have a sense of what we're trying to do with it, exactly.

ranweiler:
Also, while I'm thinking of it: with any _real-life_ example/sample data, we should ensure that the permitted uses (if any) of such data are spelled out with an equivalent of a LICENSE file.

ranweiler:
Oops, I didn't really make my motivation clear, sorry. The reason I suggest the above is so that doc authors can freely include and modify example data so that it is relevant to the text, and similarly, so software tests can be freely modified, without breaking documentation (or vice versa) This way, too, we are sure that we only track (in version control) data that is directly relevant to actual tests or explanatory documents, and avoid tracking data that _might_ be relevant, but isn't actually used for anything.

nealmcb:
@kiniry OK to pull this?

docs/glossary.md
```diff
@@ -42,19 +45,22 @@ ballot measure, that appears on the ballot for an election in a
...
-location of a ballot to allow a county worker to efficiently find the
+ballot manifest specifies the physical
+location of a ballot to allow staff to find the
 specific ballot represented by a given CVR. A ballot manifest will
 contain the following information: county ID, tabulator ID, batch ID,
-(Shown in Appendix 2.) number of ballots in each batch, and the
+the number of ballots in each batch, and the
 storage location where the batch is secured following tabulation. A
-sample ballot manifest is attached as Appendix 2.
+sample ballot manifest is provided at `test/manifest-dq.csv`.
```
ranweiler: You actually moved this to `docs/samples/`, not `test/`, right?

docs/glossary.md
```diff
@@ -42,19 +45,22 @@ ballot measure, that appears on the ballot for an election in a
...
 
-* **cast vote records (CVR)** - Electronic records created during
-scanning of paper ballots indicating how the voting system interpreted
-physical marks on every ballot. A sample CVR is attached as Appendix 1.
+* **cast vote record (CVR)** - An electronic record
+indicating how the marks on a ballot were interpreted as votes.
+May be created by a scanner or DRE, or manually during an audit.
+Sample CVRs in Dominion's format are in
+`test/dominion-2017-CVR_Export_20170310104116.csv`.
```
ranweiler: Per the earlier discussion, we should put this file in `docs/samples/` too, right?

ranweiler: That is: move the actual file, and also change this reference.

kiniry:
I'm ok with the merge on the glossary.  Note that, for future reference, I wouldn't mix different kinds of changes/contributions in the same pull request.  I'll do the conflict merge now.

docs/glossary.md
nealmcb: Oops - sorry. Yes, @ranweiler.

ranweiler:
@kiniry there is at least one erratum that should be addressed before merging, please refer to my review.

nealmcb:
@kiniry My intent was to simply make the references for the glossary concrete and add real examples, and it evolved with ranweiler's comments.  Just to be clear, are you thinking the example which I left in the top-level test directory is a "different kind of change" which  would be better as a separate PR?

kiniry:
Now that I look at the dependency between the examples and the main text, I think that you made the right call. ;)  I have merged and will delete this branch now.

PR #13, opened 2017-07-10
New build system for docs and reorg of docs to build one master document.

kiniry:
We'll eventually need to drop in some Free & Fair templates/style to make the book look great in both PDF and HTML, but for now we are only focusing on content.


nealmcb:
Doc build fine on my system.
Looks good to me!

nealmcb:
Links in master README.md to doc files need to be updated.

kiniry:
We'll fix that @nealmcb.  I'll merge this now, as I'm going to work on the glossary now.

PR #20, opened 2017-07-11
Fix README links to documentation

ranweiler:
Fixes links in README after the addition of numeric prefixes to the relevant files.


nealmcb:
Yay! +1
Do we have a tool to check the links automatically in the future?

PR #24, opened 2017-07-12
Eclipse configuration

dmzimmerman:
This pull request adds an initial Eclipse configuration, installation instructions (that have been verified to work on both Mac OS and Linux) for the Eclipse distribution with our plugins, and an Eclipse project for the server side. The Eclipse project itself is subject to changes in configuration when we start adding outside libraries, most likely via Maven.


kiniry:
I have done a test run-through of this project configuration's documentation and given @dmzimmerman feedback, which he will be committing shortly.

dmzimmerman:
The project is now a Maven project and pulls in the Sparkjava framework for server-side web; I believe it is ready for review, any further work on project setup (if it becomes necessary) can happen in another PR.

server/README-ECLIPSE.md
```diff
@@ -0,0 +1,40 @@
...
+
+The installation of the various plug-ins we use for development,
+code quality, etc., is straightforward, but is nowhere near as automated
+as we would like. To install an Eclipse distribution with all the required
+tools and get the project ready for development, do the following:
+
+  1. Download the Eclipse Oxygen installer for your platform from
+  http://www.eclipse.org/oxygen/.
+  2. Run the installer and install the "Eclipse IDE for Java Developers"
+  (the first option on the list of distributions to install).
```
nealmcb: I don't find clear or convincing installation documentation there. Do we just run `eclipse-inst`? Here's all I see.
Clicking "Download" yields a .tar.gz file.
`tar zxf eclipse-inst-linux64.tar.gz` yields an `eclipse-installer` directory.
That directory has a `readme/readme_eclipse.html` which gives no obvious hints about what to do next.
I'm guessing I should run the executable file `eclipse-inst`, but running random undocumented executables off the internet with only a grandiose sha512 hash (`2bbb38475b3f3b13b9c6c5db5895fe25ce24b4a8df6a24c6822a348a9652dc8df49ca19152495f34156c39022b8c2ed6b5bb1e84db3edc751302d73bdc1d56da  eclipse-inst-linux64.tar.gz`, FWIW) is not my idea of a secure plan.
Where is a PGP signature, or a proper signed package? Yeah, this is feedback for eclipse, but ....
This looks like a useful link: [How to install Eclipse using its installer - Ask Ubuntu](https://askubuntu.com/questions/695382/how-to-install-eclipse-using-its-installer)

dmzimmerman: Yes, you do run `eclpse-inst`, and it surprises me that the Eclipse people don't have reasonable documentation about it.

Sadly, there's no way to get "security" with an Eclipse distribution at present - even if Eclipse was signed or properly hashed, most of the static checking plugins we use are not. I don't consider it a serious problem for the security of the software under development because we don't need to build it with Eclipse (and in particular, the continuous integration build will not be performed with Eclipse). It is, of course, a more serious problem in terms of being worried about what untrusted software might do to your machine.

In the future I'd like to explore different IDEs to see if we can find one with both a better security story around it and support for all the different tools we want to use. My guess is we can't.

ranweiler:
Closes #51.

server/README-ECLIPSE.md
nealmcb: Thanks.  Understood.
Adding the askubuntu link I shared to README-ECLIPSE.md might make sense. I just haven't had time to actually run thru it yet and test this all out. But don't let that slow down merging this, if it works for someone else.

ranweiler: @nealmcb I'll open an issue for this so we capture the concern, which is valid, and also keep things movin' along!

PR #29, opened 2017-07-14
Glossary update

kiniry:
Addition of final terms to glossary gathered from various sources.  I intend to ask @mkiniry to help on this and the informal domain model next.  I'm including @mkiniry as a potential reviewer so he can start to witness the workflow of reviews in GitHub.

### DIFF
--- a/docs/89_glossary.md
+++ b/docs/89_glossary.md
@@ -228,7 +228,7 @@ non-uniformly in order to target non-county-wide contests
 
 * **county** TBD
 
-* **scanner** TBD
+* **scanner** has an ID number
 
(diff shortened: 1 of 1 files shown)


ranweiler:
@kiniry the first two commits in this branch are identical, differing only in that one is signed. Could you drop the unsigned one and force push the branch?

ranweiler:
Actually, it seems like you only _really_ have the one commit. Maybe it is best to just rebase it on `master`, dropping the unsigned and spurious merge commit.

kiniry:
I'll get around to that @ranweiler.  I'm not sure how this tiny/simple branch got into this state; probably something about the combined use of the (new to me) GitHub UI and the CLI.

PR #37, opened 2017-07-15
Define non-voter-verifiable ballot, VVPR, VVPAT; improve consistency

nealmcb:


docs/89_glossary.md
```diff
@@ -127,11 +127,25 @@ following requirements:
   * Something you know (Example: passwords)
   * Something you are (Example: biometrics)
 
+* **ENR system** - Election Night Reporting system, used to publish
+election results starting on election night, and continuing thru the end
```
ranweiler: Nit: may prefer the spelling "through" instead of "thru".

ranweiler:
I've reviewed the latest updates, and everything looks good to me!

docs/89_glossary.md
nealmcb: Fixed in 4821988. Thanks for catching that.

PR #41, opened 2017-07-17
Some minor updates to the domain model

dmzimmerman:


specs/corla.pvs
```diff
@@ -717,18 +717,18 @@ BEGIN
...
-%%% These two terms are used but underdefined at this point in time.
+%%% These two terms are used but underdefined at this time.
   audit_mistake: TYPE
   audit_progress: TYPE
 
 %%% The count of the total number of ballots to audit varies across
-%%% audit types. E.g., historically in Colorado random audits must
-%%% audit 500 ballots or 20\% of all ballots, whichever is smaller.
+%%% audit types. For example, historically in Colorado random audits
+%%% must audit 500 ballots or 20\% of all ballots, whichever is
```
nealmcb: 5%, not 20%. Sorry I missed this before....  @trace CRS 1-7-514

specs/corla.pvs
```diff
@@ -1021,11 +1020,11 @@ state_wide_dashboard: THEORY
...
 
-%%% CDOS staff, after authenticating to the state-wide dashboard, is
-%%% able to see the status of the entire election. Various static
-%%% information about the election is displayed along with dynamic
-%%% information, the type and content of which is dependent upon in
-%%% the current audit stage.
+%%% CDOS staff, after authenticating to the state-wide dashboard, can
+%%% see the status of the entire election. Various static information
+%%% about the election is displayed along with dynamic information,
+%%% the type and content of which is dependent upono the current audit
```
nealmcb: Typo: upon

specs/corla.pvs
```diff
@@ -1053,23 +1052,32 @@ BEGIN
...
-%%% CERTAIN CONTESTS; AND
-%%% (5) THE ABILITY OF THE COUNTY CLERKS TO COMPLETE THE AUDIT BEFORE
-%%% THE CANVASS DEADLINE.
+%%% No later than 5:00 PM MT on the Friday after Election Day, the
+%%% Secretary of State will select for audit at least one statewide
+%%% contest, and for each county at least one countywide contest. The
+%%% Secretary of State will select other ballot contets for audit if
+%%% in any particular election there is no statewide contest or a
+%%% countywide contest in any county. The Secretary of State will
+%%% publish a complete list of all audited contesets on the audit
```
nealmcb: typo: contests
Note: I haven't rigorously compared the language....

specs/corla.pvs
```diff
@@ -1053,23 +1052,32 @@ BEGIN
...
-%%% (3) ANY CAUSE FOR CONCERN REGARDING THE ACCURACY OF THE REPORTED
-%%% TABULATION OUTCOME OF THE CONTESTS;
-%%% (4) ANY BENEFITS THAT MAY RESULT FROM OPPORTUNISTICALLY AUDITING
-%%% CERTAIN CONTESTS; AND
-%%% (5) THE ABILITY OF THE COUNTY CLERKS TO COMPLETE THE AUDIT BEFORE
-%%% THE CANVASS DEADLINE.
+%%% No later than 5:00 PM MT on the Friday after Election Day, the
+%%% Secretary of State will select for audit at least one statewide
+%%% contest, and for each county at least one countywide contest. The
+%%% Secretary of State will select other ballot contets for audit if
```
nealmcb: typo: contests
Also, the lowercasing, while much more pleasant to read, may be inappropriate since the rule is still a draft, and thus presented in caps.
In general this makes me wonder if we want to be reformatting, at the risk of loss of fidelity.

dmzimmerman: I think this was the only section in all caps... is it the only section that is still a draft? My inclination is that readability should trump fidelity to the source material here, especially since we're not changing the wording.

nealmcb: Good point. This might be the only section that was changed from the previous draft. Readability is good, so you can just ignore this comment.

PR #43, opened 2017-07-17
Add diagram of v.1.1 of CORLA system spec.

sfsinger19103:
[CORLA System Spec v1.1.pdf](https://github.com/FreeAndFair/ColoradoRLA/files/1154030/CORLA.System.Spec.v1.1.pdf)


dmzimmerman:
Uploading the files seems fine, but I'm not sure the current "specs" directory is the right place (perhaps it itself needs to be broken into a couple of subdirectories, including one for PVS and one for other informal artifacts). Anybody else have thoughts on this?

ranweiler:
@dmzimmerman I think the formal/informal split idea seems fine. Maybe be even more specific: all PVS stuff goes in `spec/pvs`, for now, other spec artifacts just go in `spec`. How's that sound?

nealmcb:
Good questions @dmzimmermam, and this is a good time for us to consider them.
The diagram might fit better in to the doc folder.
I'd also prefer that we avoid spaces in file names, though I haven't checked to see if there are any others already in master.
I'd generally like to distinguish source files (the .graffle) from derived files (the pdf), perhaps in a subdirectory for `build` products or some other common term that fits with the way we'll build the latex documentation.

nealmcb:
@ranweiler I'm not sure I know enough about how all the various specs will relate. E.g. will the pvs work ever produce flows? Is there any formal spec (code of some sort which documents them and from which graphics could be derived) for the screens and flows, or will it be hand-made images, in which case the specs directory is probably the right place. A subdirectory for pvs sounds appropriate offhand in that case.

ranweiler:
@nealmcb I would suggest that we choose an organization scheme that seems fine and low overhead, and change it as needed, as we add more artifacts. Whatever we do in this PR needn't be permanent, just enough to make sure no toes get stepped on as we proceed on the specs themselves.

dmzimmerman:
I'm happy enough moving the PVS stuff to a "pvs" directory, and leaving these two artifacts here. I also agree with Neal that we should try to avoid filenames with spaces to the extent possible (a quick `find . -name "* *"` reveals that there are currently none in the repo).

ranweiler:
Now that #45 has landed, I am going to rebase this branch on `master` and merge it.

PR #45, opened 2017-07-18
Move the PVS files and clean up some of the LaTeX therein

dmzimmerman:


specs/pvs/corla.pvs
```diff
@@ -713,8 +719,8 @@ BEGIN
 %%% constituancies and have various roles. Deciding who is on an audit
```
ranweiler: Unrelated typo: should be "constituencies".

PR #64, opened 2017-07-19
Updates to the server Eclipse project and some initial boilerplate code

dmzimmerman:


dmzimmerman:
Well. It seems that CI works for the server project, but right now it's only doing the CheckStyle and PMD checks (and the build). Unfortunately I don't have time to integrate more of the checking at the moment, and I'm also not sure the way I've done it is the right way to do it going forward (it may be better to write a single shell script that does the checks and build, or to make the Maven build file do the checks as an integral part of the build instead of invoking them separately).

ranweiler:
@dmzimmerman this looks good for a first pass on the CI! Did you verify that each check can fail, too?

dmzimmerman:
I verified on my local machine that the PMD check can fail. The Checkstyle check will report errors to the console but somehow still let the "build" succeed, and I have no idea why that is - it's probably mainly due to my relative lack of experience with Maven, in combination with something odd about the way the Checkstyle plugin works. I'll look into it some more at some point.

ranweiler:
Er, to be clear, I mean "fail" in the Unix-observable sense of a nonzero exit code.

dmzimmerman:
Yes, I got that.

ranweiler:
Ah, I see. We need to make the checks actually return nonzero if it fails, or else it isn't actually integrated into the CI reporting and merge restrictions.

dmzimmerman:
Yes, I'm aware of that. So I'll keep looking at it perhaps later tonight and try to figure out how to make it do what we want it to do.

dmzimmerman:
I think it's working now (at least for CheckStyle and PMD; I haven't put the FindBugs plugin in there yet).

server/eclipse-project/src/main/resources/us/freeandfair/corla/default.properties
```diff
@@ -0,0 +1,5 @@
+# The default properties file for ColoradoRLA
+port = 8888
+locale = en_US
+keystore = /us/freeandfair/corla/default.jks
+keystore_password = corla2017
```
dmzimmerman: Well, it's the default properties file (that is, the "if you check out the project and build it without configuration, this is what you run" file). In production, there will be a properties file that actually points to a legit keystore (which, of course, cannot be stored in the repo) and has a legit keystore password. And other properties that have not yet been added there, of course, like "database location".

PR #72, opened 2017-07-20
Fix Eclipse setup to work with plugin updates

dmzimmerman:
Some plugin update sites don't serve old versions; when that happened with CheckStyle today, it broke the setup. So, I've removed the maximum version number restrictions from the setup file.

### DIFF
--- a/server/eclipse.setup
+++ b/server/eclipse.setup
@@ -109,7 +109,7 @@
       licenseConfirmationDisabled="true">
     <requirement
         name="AGFeature.feature.group"
-        versionRange="[0.3.1,0.4.0)"/>
+        versionRange="0.3.1"/>
     <repository
         url="http://kindsoftware.com/products/opensource/AutoGrader/update/"/>
   </setupTask>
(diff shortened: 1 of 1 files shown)


ranweiler:
@dmzimmerman, this is question is tangential, out of curosity, does Eclipse have any concept of lock files for resolved dependency sets (+constraints)? I'm thinking of something analogous to a [`Cargo.lock`](http://doc.crates.io/guide.html#cargotoml-vs-cargolock) from Rust, or an [`npm-shrinkwrap.json`](https://docs.npmjs.com/files/shrinkwrap.json) from NPM/Node.

dmzimmerman:
To the best of my knowledge, not exactly... The problem is that you can come up with such a file - which we pretty much had with that setup file - but if the maintainers of the update sites don't keep old versions around, there's nowhere to get them from.

ranweiler:
> To the best of my knowledge, not exactly...

I see. Like you said, wouldn't help in this case anyway, since there's the scattered dependency sites that might not support a known good lockfile.

dmzimmerman:
No, it's removing the maximum version number limit; so it's enforcing a minimum version but later ones are OK. I am not particularly worried about the consequences of some future plugin update breaking something, because (1) the worst case scenario there is a small amount of reconfiguration of static checks would be needed, (2) Eclipse is not used to actually build the product, and (3) historically, other than one very major change in CheckStyle's API, there have been very few changes to these tools that have broken our Eclipse configurations (in my decade of using them, I've had to revise one set of configuration files once for compatibility reasons).

ranweiler:
> No, it's removing the maximum version number limit;

Ah, I see: https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

So `"v1"` is the same as `"[v1,)"`.

PR #76, opened 2017-07-20
Refined some static analysis and HTTP/HTTPS listening

dmzimmerman:


server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
```diff
@@ -75,29 +97,70 @@
...
-      LOGGER.info("could not read port number from properties, using default port");
+      LOGGER.info("could not read port number property " + the_property + 
+                  ", using default " + the_default);
+    }
+    
+    return result;
+  }
+  
+  /**
+   * Redirect a request from HTTP to HTTPS.
```
ranweiler: To me, the existence of this method is evidence that it would be better to configure TLS at Nginx, terminate it there, and reverse proxy. Failing that, I would suggest never serving plain HTTP at all.

server/eclipse-project/src/main/resources/us/freeandfair/corla/default.properties
```diff
@@ -1,5 +1,6 @@
 # The default properties file for ColoradoRLA
-port = 8888
+http_port = 8888
+#https_port = 8889
```
ranweiler: Are this property and `keystore` both mean to be commented out?

dmzimmerman: Oops. No, they are not. I will fix that and re-commit in a bit.

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
dmzimmerman: Indeed; I wanted to make it so that it could run standalone if necessary, and it was a pretty straightforward thing to do.

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
```diff
@@ -80,24 +87,65 @@ public Main(final /*@ non_null */ Properties the_properties) {
...
+   * 
+   * @param the_request The request.
+   * @param the_response The response.
+   * @param the_port The HTTPS port.
+   */
+  private void httpsRedirect(final Request the_request, final Response the_response,
```
ranweiler: The body of this method is a fairly straightforward pure function `String -> String`, with some obvious possible edge cases. We could factor it out and add a few unit tests.

Alternately, we could use the Java `URL` class to parse the `String` we get from `request.url()`, then use the class again to construct a `URL` from exactly the parsed parts, modifying exactly the protocol and port, converting back to a `String` at the end. That way we can lean on the standard lib instead of implementing custom parsing.

I prefer the latter approach, since it will result in code that should be fairly obviously correct. What do you think?

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
kiniry: We'll make sure to discuss this with the IT guys in CDOS tomorrow morning.

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
kiniry: I am aligned with that implementation style.

dmzimmerman: I thought about doing the latter, and even started writing it... but the URL class is immutable, and none of the URL constructors takes sufficient parameters for that to work universally (i.e., if the URL has a query or an anchor component - admittedly unlikely cases for this system - you can't pass those to a new URL constructor and they would be lost).

ranweiler: Ugh, true. However, it looks like we could use a [URI](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#URI(java.lang.String,%20java.lang.String,%20java.lang.String,%20int,%20java.lang.String,%20java.lang.String,%20java.lang.String)), then from that obtain a [URL](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#toURL()), as long as those docs aren't too new.

dmzimmerman: Actually, it looks like `URL.getFile()` does more than one would think it should, and returns the query string as well. So I can probably replace it with something like 

```URL newURL = newURL("https", oldURL.getHost(), the_port, old.getFile());```

I'll give that a try.

ranweiler: (Or maybe just the `URI.toString()` would be fine?)

dmzimmerman: And... it actually worked, much to my surprise. You'd really think that `getFile()` would mean to get the _file_, but it gets basically the entire trailing part of the URL, queries and anchors and all.

dmzimmerman: So much for intuitively-named methods.

ranweiler: Never know when you're gonna run into a 100 GB URI query...

dmzimmerman: Well, if that happens, we either don't even get this far (Spark/Jetty chokes and dies on the query string) or it gets sanitized before we do.

ranweiler: Ha, sorry, forgot the ":stuck_out_tongue_winking_eye:"

Issue #79, opened 2017-07-21
data model for server-side architecture


kiniry:
Define the data model of the server-side architecture, derived from the UI model and underlying system requirements.

kiniry:
The initial data model is complete and a first implementation of it is finished as well.  @dmzimmerman is revising it to work with Hibernate now in the `jpa-work` branch. https://github.com/FreeAndFair/ColoradoRLA/tree/jpa-work

PR #80, opened 2017-07-21
Data model

kiniry:
I'm working on the data model, per issue #79.  I hope to complete it tonight so that it is ready for review tomorrow (Friday).


specs/pvs/corla.pvs
```diff
@@ -1650,6 +1761,186 @@ END system_assumptions
...
+    sequence[[political_party]]
+  national_contest_table: TYPE =
+    sequence[[nation, contest]]
+  state_contest_table: TYPE =
+    sequence[[state, contest]]
+  county_contest_table: TYPE =
+    sequence[[county, contest]]
+
+  public_meeting_to_determine_seed_table: TYPE =
+    [# location: string, date: string #]
```
dmzimmerman: Is this meant to be a single record or a sequence of records? More generally, how are you deciding where ordering of the fields in the individual "rows" is important (`[[ ... ]]`) and where it is not (`[[# ... #]]`)?

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1761,186 @@ END system_assumptions
...
+  cvr_table: TYPE =
+    sequence[[county, cvr]]
+
+%%% But we also need to store all CVRs for each contest, aggregating
+%%% all CVRs across the upload from all counties in which the contest
+%%% was on the ballot.
+
+  contest_cvr_table: TYPE =
+    sequence[[contest, cvr]]
+
```
dmzimmerman: For the above two CVR tables, are we storing the individual CVRs redundantly in both tables, or is the idea that there is a single table containing CVRs and some ID assigned to each, which is then referenced in both these tables?

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1761,186 @@ END system_assumptions
...
+  audit_reports_table: TYPE =
+    sequence[[county, audit_report]]
+  rla_summary_results_table: TYPE =
+    sequence[[contest, summary_results_report]]
+
+%%% All of the information on the public dashboard is derived from the
+%%% above data.
+
+% @todo kiniry Add election as first column to every table? Or have a
+% different DB per election?
```
dmzimmerman: Either is likely to work fine given the data volume we're talking about here. Having a separate DB per election has appeal because of the otherwise unbounded growth of those tables over time, and for ease of archiving.

specs/pvs/corla.pvs
kiniry: There is only a single meeting per election, so I didn't make this a sequence.  If we decide to put multiple elections in the same DB, then this will get lifted to `[election, [# ... #]]`. In general, I'm modeling everything as a tuple unless there is ambiguity in projection type semantics, as is the case here since this is a string x string.  I'll probably migrate all tuples to records at some point, but that's not a priority.

specs/pvs/corla.pvs
kiniry: The latter.  The idea is to only put them in one table and to use a relation to map them into other tables.  As I mentioned this morning face-to-face, I realized last night as I went to bed I had forgotten to explicitly specify the relations like this that span tables.  I'll do that today.

specs/pvs/corla.pvs
kiniry: I'm leaning that way too.  A middle ground is to do one DB per state per year, e.g.

specs/pvs/corla.pvs
nealmcb: I'll just note as I did re: CollectionID vs CountyID above, that there will be times where counties will want to add more ballots after the random seed is selected. If they want to include those in an audit, there will then have to be another random seed generation ceremony.  But hopefully not in November 2017, since we don't have well-documented algorithms (e.g. stratified Bayes audit) to specifically deal with it yet....

specs/pvs/corla.pvs
```diff
@@ -254,6 +271,9 @@ BEGIN
   full_hand_count_contest: TYPE FROM contest
 % @todo Contest name uniqueness.
 
+% @todo kiniry These types will be refined when we review the
+% information provided by CDOS on 20 July 2017 about Dominion's file
+% formats.
   outcome: TYPE
```
sfsinger19103: renaming "outcome" to "contest_outcome" might be clearer

specs/pvs/corla.pvs
```diff
@@ -347,7 +377,10 @@ END election
 % @relation box inherit storage_container
```
sfsinger19103: Line 339/369: Isn't a "ballot" a record of voter intent, rather than a "means"?

specs/pvs/corla.pvs
```diff
@@ -372,8 +405,8 @@ BEGIN
 % @todo Each scanner has a unique id number.
   vvpat: TYPE
```
sfsinger19103: Is "VVPAT" the official Colorado terminology?

specs/pvs/corla.pvs
```diff
@@ -372,8 +405,8 @@ BEGIN
...
   verified?: TYPE = pred[storage_container]
-  log: TYPE
-  chain_of_custody_log: TYPE FROM log
+
+  vvsg_log: TYPE
```
sfsinger19103: Are we really talking about Voluntary Voting Systems Guidelines here? If so, please explain. If not, please change this name.

specs/pvs/corla.pvs
```diff
@@ -243,8 +258,10 @@ BEGIN
 %%% unique name.
 % @review CDOS will there be one risk limit? If more than one, then
 % system must provide a way to map contests to risk limits.
```
sfsinger19103: There will be not one risk limit but some number of (Contest, risk limit) pairs.

specs/pvs/corla.pvs
```diff
@@ -422,6 +455,7 @@ BEGIN
   original_ballot: TYPE FROM ballot
```
sfsinger19103: Line 405/438 ff: it's unclear what we mean, precisely, by ballot. Technically, I doubt we want a single instance of a ballot to refer both to a paper ballot and a digital representation. Needs correction or, if correct, clarification.

specs/pvs/corla.pvs
```diff
@@ -422,6 +455,7 @@ BEGIN
...
   duplicated_ballot: TYPE FROM ballot
   non_voter_verifiable_ballot: TYPE FROM ballot
+  voter_verifiable_ballot: TYPE FROM ballot
   phantom_ballot: TYPE FROM ballot
 
 %%% Ballots can be in various stages of processing, as implicitly
```
sfsinger19103: Line 428/262, term "C.R.S." not defined

specs/pvs/corla.pvs
```diff
@@ -438,8 +472,10 @@ BEGIN
 
 % @description A decision to be made, either the selection of one or
 % more candidates for an office or the choice of a course of action.
+% @todo kiniry Need to clarify choices vs options vs votes.
   ballot_contest: TYPE = contest
```
sfsinger19103: What thing of type "contest" would not be a "ballot_contest"? (Or perhaps I've misunderstood the syntax and this is just saying that "contest" and "ballot_contest" are synonyms?

specs/pvs/corla.pvs
```diff
@@ -486,11 +527,15 @@ END ballot_manifests
 % @relation well_formed_vote inherit vote
```
sfsinger19103: line 476/517  what does this line correspond to in the real world (verified?: pred[ballot_manifest])? In what sense is a ballot manifest verified or not verified?

specs/pvs/corla.pvs
```diff
@@ -541,16 +586,27 @@ END ballots_collections
...
 BEGIN
-  IMPORTING information_systems
+  IMPORTING information_systems, ballot_manifests, election, audits
   
-  report: TYPE
+  report: TYPE = [# name: string #]
 % @review CDOS The contents of this report are underspecified at the
 % moment. See submit_audit_mistake_report below for more information.
-  audit_mistake_report: TYPE FROM report
+  audit_mistake_report: TYPE =
```
sfsinger19103: "Mistake" is judgmental. Maybe better to use "discrepancy"? Audit board could be wrong in its interpretation.

specs/pvs/corla.pvs
```diff
@@ -1053,9 +1136,9 @@ BEGIN
 %%% ballot polling audits, but in no event will the risk limit exceed
```
sfsinger19103: why mention ballot polling audits in the spec, since the system doesn't support ballot polling audits? If we're mentioning them for readers such as CORLA2 or SAWG members, we could put them into some bit of prose that's not buried in the spec details.

specs/pvs/corla.pvs
```diff
@@ -725,7 +793,10 @@ BEGIN
 
 % @review CDOS Must audit boards have a minimum number of members?
 % A maximum?
+% @trace Per email from Jerome Lovato at CDOS on 19 July 2017: The
+% audit board has exactly two members.
   audit_board: TYPE = set[audit_board_member]
+  audit_board_size: AXIOM TBD
```
sfsinger19103: CDOS has told us that the size is 2

specs/pvs/corla.pvs
```diff
@@ -725,7 +793,10 @@ BEGIN
...
 
 %%% These two terms are used but underdefined at this time.
```
sfsinger19103: When will they be defined? CDOS reader will want to know.

specs/pvs/corla.pvs
```diff
@@ -438,8 +472,10 @@ BEGIN
...
 % @description A decision to be made, either the selection of one or
 % more candidates for an office or the choice of a course of action.
+% @todo kiniry Need to clarify choices vs options vs votes.
   ballot_contest: TYPE = contest
-  choice: TYPE
+  option: TYPE = string
+  choice: TYPE = set[option]
 
 % @description A batch is a set of ballots.  
   batch: TYPE = set[ballot]
```
sfsinger19103: line 452/488 meaning of "imprinted ballot" should be made clear in prose. I can't do it since I don't know what "imprinted ballot" means.

specs/pvs/corla.pvs
```diff
@@ -863,6 +934,8 @@ BEGIN
 % @description The audited contest determines the number of ballots
```
sfsinger19103: line 812/883 For example, we *could* assume

(add "could")

specs/pvs/corla.pvs
```diff
@@ -863,6 +934,8 @@ BEGIN
...
 % that must be examined and verified during the RLA.
   number_of_ballots_to_audit: [audited_contest -> nat]
+
+% @todo kiniry Still need to model core RLA algorithm(s).
 END rlas
 
 %+END audits
```
sfsinger19103: line 878/971 ff: 
"They frequently, but not always, contain an interpretation of all contests on a ballot."

better might be: "an interpretation of voter choices in all contests on a ballot"

sfsinger19103: line 896/969 what real world verification of a cvr does this line model? I.e., what does it mean for a cvr to be verified or unverified?

specs/pvs/corla.pvs
```diff
@@ -922,8 +995,10 @@ END cast_vote_records
 %%% the county-level, and one is for audit board members.
```
sfsinger19103: Incorrect!! Audit Board members use the County interface. And should the public dashboard be mentioned here? Is it a "user interface"?

specs/pvs/corla.pvs
```diff
@@ -1031,7 +1113,8 @@ END dashboard
 % @trace Principly RDPR-6-Jul-2017 Section 25.2.2
```
sfsinger19103: line 1023/1105 change to "Department of State Dashboard", as "State-wide Dashboard" might be mistaken for the public dash (which is also state-wide).

specs/pvs/corla.pvs
```diff
@@ -1133,6 +1216,13 @@ BEGIN
   random_list_of_ballots:
```
sfsinger19103: line 1113/1196 The system we are delivering will use only ballot_permutation, as far as I know. So we should not include the other here. Why confuse the reader?

specs/pvs/corla.pvs
```diff
@@ -1133,6 +1216,13 @@ BEGIN
...
+% @trace Per email from Jerome Lovato at CDOS on 19 July 2017: The
+% tool must also provide a printable list of ballot for each round.
+% At a minimum, the list must contain ballot style and ballot manifest
+% information.
+  print_ballots_under_audit_list:
+    [sequence[ballot] -> ballots_under_audit_instructions]
+
 % @trace RDPR-6-Jul-2017 Section 25.2.2(K)
 
 %%% \subsection{Random Selection of Ballots for Audit}
```
sfsinger19103: line 1141/1231 This sounds as if SoS needs to provide a system for pseudo-random number generation, when in fact the RLA tool provides the pseudo-random number generation. All the SoS provides is the seed. Clarify.

specs/pvs/corla.pvs
dmzimmerman: Yes, it is saying that `contest` and `ballot_contest` are synonyms.

specs/pvs/corla.pvs
dmzimmerman: As I understand it, this report is actually about mistakes, i.e., erroneous user actions that get logged, not about audit discrepancies.

specs/pvs/corla.pvs
dmzimmerman: Isn't that part of the spec a verbatim reproduction of RDPR-6-Jul-2017 Section 25.2.2(A)?

specs/pvs/corla.pvs
dmzimmerman: @aisamanra? Do you mean @ranweiler or Morgan? To the best of my knowledge, Getty Ritter is not working on this project.

specs/pvs/corla.pvs
dmzimmerman: Again, this is a direct reproduction of RDPR-6-Jul-2017 Section 25.2.2(K).

specs/pvs/corla.pvs
ranweiler: It may be appropriate to describe the public dashboard here. However, unlike the other interfaces here, it is only "interactive" in a superficial way: public users cannot modify any data in the RLA system, they are just presented with read-only views of the data. It is still fine to specify _what_ they are presented with, but @kiniry may prefer to spec that elsewhere?

specs/pvs/corla.pvs
nealmcb: I agree that we shouldn't call it an election outcome.  We are doing the audit contest-by-contest, and leaving some out, so we don't have an outcome for the election.

I'm presuming that the glossary is backing up this document, and that should probably be noted at the top. I guess we might also try to use convention to explicitly or implicitly refer to it in cases like this.

The glossary defines outcome in terms of a set of contest winners.  This actually gets trickier, since some elections have different classes of winner (e.g. Boulder city council 5 at-large seats, last-place winner gets a two-year term, others get 4-year terms). But I doubt that information will be discernable from the CVR file. For now I'd list that as a thorny issue that we may not get to this time around.

It looks like election_outcome, as a set of outcomes, might be this set of winners, in which case I'd suggest terms like `winner`, and `outcome`, to replace `outcome` and `election_outcome`. (Or, someday, an outcome could be a set of sets for each class_of_winners).

specs/pvs/corla.pvs
```diff
@@ -6,7 +6,18 @@
...
+%%% itself will guide deployment, management, maintenance and evolution 
+%%% of the system delivered to the Colorado Department of State. The prose
+%%% annotations should help Colorado Department of State elections domain experts
+%%% and technical staff to assess the completeness and correctness of the 
+%%% ColoradoRLA system/tool.
+%%% 
+%%% This document is written in a literate style using the PVS
 %%% theorem prover. Writing system specifications in this fashion is
 %%% called (formal) domain engineering. Our formal domain model is
 %%% written in PVS's higher-order logic (HOL). Our informal domain
```
sfsinger19103: line 153/164 revise to get rid of the "?" placeholder for missing information

specs/pvs/corla.pvs
nealmcb: This comment somehow got disassociated from the line in question, which seems to be 
`%%% the ballot. Ballots are the means by which voters indicate their`
I think the term "means" helps deal with the fact that some types of ballot defined here, like duplicated_ballot are intermediate forms, which may or may not be a record of voter intent.

specs/pvs/corla.pvs
sfsinger19103: line 1194/1284 I believe there is a requirement that in each County the two Audit Board members must be of different parties.

specs/pvs/corla.pvs
nealmcb: Good catch. Note that since this is a comment on text that isn't near a revision from this PR, we can't easily link the comment to the line.  And since the line number can change with future commits, here is some more context, lines 163-166:

```
%%% values. Email is used for communication between officials and
%%% with the ?. Files are used to store information in a
%%% persistent fashion. Consequently, we introduce a few types for
%%% these ideas.
```

specs/pvs/corla.pvs
```diff
@@ -1309,9 +1399,29 @@ BEGIN
...
+     list[ballot_mark] -> audit_board_dashboard]
+
+% @trace Per email from Jerome Lovato at CDOS on 19 July 2017: If the
+% Audit Board cannot find a ballot, the tool should have a “ballot not
+% found” box that the Audit Board will select. This will be marked as
+% an error and the county may have to audit additional ballots.
+  report_ballot_not_found:
+    [audit_board_dashboard, phantom_ballot -> audit_board_dashboard]  
+
+%%% If a ballot does not have a voter-verifiable paper ballot associated
```
sfsinger19103: I've asked Dwight for suggestions of language familiar to CDOS and Counties (and am waiting on a reply, as of 2:30pm PDT Friday). He told me earlier that "VVPB" is not a widely recognized acronym. So we may wish to change this language.

specs/pvs/corla.pvs
```diff
@@ -1309,9 +1399,29 @@ BEGIN
...
+%%% with it then we report such and recompute.
```
sfsinger19103: Put Audit Board in the driver's seat:  "then the Audit Board reports the lack of voter-verifiable paper ballot."

sfsinger19103: line 1327/1437 capitalize Secretary of State

specs/pvs/corla.pvs
```diff
@@ -1309,9 +1399,29 @@ BEGIN
...
+  report_ballot_has_no_voter_verifiable_paper_record:
+    [audit_board_dashboard, non_voter_verifiable_ballot ->
+     audit_board_dashboard]  
 
 %%% If supported by the county’s voting system, the audit board may
 %%% refer to the digital image of the audited ballot captured by the
```
sfsinger19103: line 1338/1448 "Continue until the risk limit **for each** audited contest..."

specs/pvs/corla.pvs
nealmcb: Upon further reflection, I'm thinking `non_voter_verifiable_ballot` should be tracked and reported on, but should not move an audit towards a full hand count. 
This is based on a query from @mkiniry and recent calculations and discussion with CORLA2 (where we may not have unanimity on this). We don't trust the `non_voter_verifiable_ballot`, but we've got nothing better. Trigering escalation would be pointless because a full hand count can't correct anything with respect to these ballots, and the effort would be wasted.

We could just say those ballots should not be included in an audit at all ("don't waste your time on them"), and attach a 
caveat to the final report documenting the number of ballots that were assumed to be correct, but that sounds like an out-of-scope  enhancement request to me at the moment.

specs/pvs/corla.pvs
```diff
@@ -1499,7 +1619,9 @@ END authentication
 
 % @relation back_end inherit system_architecture_component
 % @relation front_end inherit system_architecture_component
-system_architecture_component: DATATYPE WITH SUBTYPES back_end, front_end
+system_architecture_component: DATATYPE
+  WITH SUBTYPES back_end,
+                front_end
 BEGIN
   IMPORTING information_systems, database, server
```
sfsinger19103: line 1578/1700 "Minimum cardinality" -- for readability, rewrite: "At least two of each are necessary because..."

sfsinger19103: line 1585/1707 Maybe replace "entails having" with "need"? Or some other rewrite for clarity.

sfsinger19103: and substitute "among" for "amongst"

sfsinger19103: line 1589/1711 typo: replace "certianly" with "certainly"

sfsinger19103: line 1635/1757 "score" should be all caps "SCORE". We should be clear that the RLA tool in no way helps the Counties with the SCORE consistency check. Nor the other checks mentioned in this section if I understand correctly. So why are we including this information in this document?

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1772,186 @@ END system_assumptions
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+%%% \section{System Logging}
+
+%%% The system must log a variety of events. One reason to log
+%%% information is to help understand how the system is operating, fix
+%%% bugs post-facto, understand how users are using the system,
+%%% etc. Another, more important reason to log in this context of this
+%%% system is to provide an indeliable record of administrator and
```
sfsinger19103: indelible (typo)

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1772,186 @@ END system_assumptions
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+%%% \section{System Logging}
+
+%%% The system must log a variety of events. One reason to log
+%%% information is to help understand how the system is operating, fix
+%%% bugs post-facto, understand how users are using the system,
+%%% etc. Another, more important reason to log in this context of this
```
sfsinger19103: strike ", more important"

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1772,186 @@ END system_assumptions
...
+%%% data model of the system, we need to:
+%%% \begin{enumerate}
+%%% \item identify the \emph{core concepts} that must be persistent in
+%%% the system
+%%% \item identify the \emph{relationships} that must be persistent in the
+%%% system
+%%% \item derive from these pieces of information:
+%%%   \begin{enumerate}
+%%%   \item the necessary \emph{tables} that \emph{collect concepts},
+%%%   \item identify the \emph{indexing concept} of each table, and
```
sfsinger19103: strike "identify"

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1772,186 @@ END system_assumptions
...
+%%%   \item define the appropriate \emph{joins} that realize
```
sfsinger19103: replace "define" with "definitions of"

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1772,186 @@ END system_assumptions
...
+% @todo kiniry Add election as first column to every table? Or have a
+% different DB per election?
+
+END data_model
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
 %%% \section{Domain Model}
 
 %%% Thus, the overall domain model of this system is the sum total of
```
sfsinger19103: Replace "Thus, the" with "The

specs/pvs/corla.pvs
nealmcb: I suggest we change the term "mistake" to "investigation".

If the audit board enters an interpretation of a sampled ballot which is discrepant with the CVR, there are a number of possibilities to investigate. Some represent discrepancies to be handled via the RLA math, and others might simply indicate that the wrong ballot was retrieved, and they may be looking at the wrong paper ballot. I'll flesh these out.

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1772,186 @@ END system_assumptions
...
+  geography_table: TYPE =
+    sequence[[nation, state, county]]
+  political_party_table: TYPE =
+    sequence[[political_party]]
+  national_contest_table: TYPE =
+    sequence[[nation, contest]]
+  state_contest_table: TYPE =
+    sequence[[state, contest]]
+  county_contest_table: TYPE =
+    sequence[[county, contest]]
```
nealmcb: Why are these 3 types of contest stored separately, rather than having them be types of contest? What about contests in districts that overlap in different ways like congressional, legislative etc?

docs/glossary_concepts.txt
```diff
@@ -0,0 +1,177 @@
...
+* **opportunistic audit** - An auditing technique designed to efficiently
+* **opportunistic contest** - A contest to be audited opportunistically.
+* **original ballot** TBD
+* **other personnel** - Individuals and Subcontractors, in addition to Key
+* **overstatement of the margin** TBD
+* **overvote** TBD
+* **paper ballot** TBD
+* **password** TBD
+* **penetration testing** TBD
+* **phantom ballot** A ballot which is listed in the manifest, but is
```
MorganMillerUX: Is this the same thing as a ballot which cannot be located but was counted in the election?

specs/pvs/corla.pvs
```diff
@@ -1650,6 +1772,186 @@ END system_assumptions
...
+  chain_of_custody_log: TYPE FROM log
+% @trace Per email from Jerome Lovato at CDOS on 19 July 2017: Logging
+% of Audit Board and SOS user errors during use of the system [...is
+% the only logging we care about].
+  rla_tool_log: TYPE FROM log
+END logging
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+%%% \section{Data Model}
```
nealmcb: For reference, here's my Django code for modeling the database in `audit_cvrs`, which I used to track CVR selections, status and discrepancies in several of the Colorado pilot RLAs. Note it doesn't (yet) have logic to store contest selections in the database or automatically compare the selected CVR to the auditors interpretation of the paper ballot.
https://github.com/nealmcb/audit_cvrs/blob/master/audit_cvrs/models.py

nealmcb: And here's the Django code for batch-level comparison auditing from electionaudits, used to run the Boulder RLA in 2010, and the more extensive almost-RLA in 2008. 
 Classes for CountyElection, Contest, Batch, ContestBatch, Choice, VoteCount, Margin http://bazaar.launchpad.net/~nealmcb/electionaudits/trunk/view/head:/electionaudits/models.py

nealmcb: A picture of a more recent data model (which I don't think I've actually put online....), with HandContestBatch and HandVoteCount classes is visible in the doc/model_graph.png file, e.g. here: model_graph

specs/pvs/corla.pvs
kiniry: I've refined the `county` definition and lifted `county_id` to a new type and included it in the `ballot_manifest_info`.  The current design, as well as the UI mock-up from CDOS, includes no ability to add ad hoc ballots.  It isn't clear to me what you mean by needing a new seed if they want to add new ballots, since the original seed determines an order.  Perhaps you can bring that up on Slack to discuss.

specs/pvs/corla.pvs
```diff
@@ -310,15 +319,25 @@ END elections
 %%% national and local elections; etc. This RLA Tool focuses on state
 %%% elections in Colorado, so we only model that one kind of
 %%% election. Several other kinds of elections are mentioned as well.
+%%%
+%%% Note that county elections are actually \emph{multi}-county
+%%% elections, where the typical case is that the set of counties is
+%%% a singleton.
```
kiniry: I've augmented the description to include your characterization.

specs/pvs/corla.pvs
nealmcb: Correct - I should have marked my comment as being related to possible future enhancements. This would be a way to actually audit `phantom ballots` that showed up late. If they do want to add more ballots, and do random selections of them, the only way to do that is to then publish CVRs for those new "collections" of ballots, generate a new seed, and figure out how to combine the results of the original audit with results from the new one. Messy, but sometimes things happen.....

specs/pvs/corla.pvs
```diff
@@ -438,8 +461,10 @@ BEGIN
 
 % @description A decision to be made, either the selection of one or
 % more candidates for an office or the choice of a course of action.
+% @todo kiniry Need to clarify choices vs options vs votes.
   ballot_contest: TYPE = contest
-  choice: TYPE
+  option: TYPE = string
+  choice: TYPE = set[option]
```
kiniry: Right now option is abstracted to a string, which can, of course, encode anything.  Right now I have encoded a voter choice as a set of options, thus this model only handles standard multi-seat plurality voting schemes.

specs/pvs/corla.pvs
```diff
@@ -470,6 +495,11 @@ ballot_manifests: THEORY
...
   IMPORTING ballots, elections_equipment, information_systems
 
+% @todo kiniry Should this information also include what kind of
+% ballot each ballot is?  I.e., re-marked, phantom, etc.?
+% @todo kiniry These types will be refined when we review the
+% information provided by CDOS on 20 July 2017 about Dominion's file
+% formats.
   ballot_manifest_info: TYPE =
     [scanner_id_number, batch_number, batch_size, storage_container]
   ballot_manifest: TYPE = set[ballot_manifest_info]
```
kiniry: I've added `county_id` for the moment.  I'm not going to get ahead of ourselves with the design to include data model elements necessary for stratified audits.  We can/will put out a new version that will handle other audit styles, and of course the data model will evolve with those new releases.

specs/pvs/corla.pvs
```diff
@@ -575,6 +620,13 @@ BEGIN
   instructions: TYPE
   ballot_instructions: TYPE FROM instructions
   sos_voter_intent_guide: TYPE FROM instructions
+
+% @trace Per email from Jerome Lovato at CDOS on 19 July 2017: The
+% tool must also provide a printable list of ballot for each round.
```
kiniry: Corrected.

specs/pvs/corla.pvs
kiniry: Got it.  ? becomes "public" for now.

specs/pvs/corla.pvs
kiniry: Yes, I'll remove that @review and other ones that have now been addressed by CDOS.

specs/pvs/corla.pvs
kiniry: I'll update the preface wrt the glossary.  As mentioned on Slack, @mkiniry is double-checking the relation between the glossary terms and the specification.  We talked over the notion of nouns, adjectives/adverbs, and verbs being types, predicate types, and functions, and the means by which that we will update the specification with @description and @explanation annotations so that the glossary becomes an automatically generated document.

I'm not going to dive into those crazy classes of contest_outcomes for the moment.  Let's just add a @note annotation that reminds us to return to that matter when necessary.

Note that `contest_outcome` is currently under-specified so as to leave open the possibility of encoding these subtleties as necessary for this release.  E.g., a legitimate encoding right now could simply be something like `[# winner: option #]`.  I didn't intend to refine this until we go over the Dominion artifacts.

specs/pvs/corla.pvs
kiniry: Yes, @nealmcb captured my intention.  The general idea of a ballot is that it convey's the meaning of a voter's intent, but by no means is it actually a record of voter intent, since so many kinds of ballots are not voter-verifiable.

specs/pvs/corla.pvs
kiniry: We don't know.  Given recent comments from CDOS, it sounds like they have some local terminology we need to learn about VV* in general.  @nealmcb, can you give us a brief tutorial on such on Slack?

specs/pvs/corla.pvs
kiniry: Yes, I am talking about the kinds of logs defined in the VVSG.  I'll expand and explain.

specs/pvs/corla.pvs
kiniry: All of the different flavors of ballot mentioned in the literature are all subtypes of `ballot`.  E.g., `digital_ballot` and `paper_ballot`.  The current specification does not denote if any of those subtypes are identical to, or disjoint from, any other.  If/when CDOS or CO law dictates such relationships, we will encode them in the specification and the software.

A bit of background for clarity: PVS's notion of type and subtype are using the types-as-sets interpretation.  So for each type you see declared, you should think of it as a set with a classifier (in the standard set theoretic sense).  Thus a subtype is nothing more than a subset classifier.

specs/pvs/corla.pvs
kiniry: I've added a @todo to the top to drop in some background material:

``` @todo kiniry Define C.R.S. and explain the other sources used to derive all concepts in this specification.```

File: specs/pvs/corla.pvs (hunk 17 above)
Comment (kiniry) [inline specs/pvs/corla.pvs]: I've added expository text.

File: specs/pvs/corla.pvs (hunk 39 above)
Comment (kiniry) [inline specs/pvs/corla.pvs]: Note that I'm leaving in this comment for the moment since I still do no think we have clarity on the contents of Dominion's ballot manifest.

File: specs/pvs/corla.pvs (hunk 12 above)
Comment (kiniry) [inline specs/pvs/corla.pvs]: In the C.R.S. the adjective "verified" is used, thus there is a notion of "verified" ballot manifest and "unverified" ballot manifest.  Anytime you have an adjective or adverb modifier that encodes a decidable truth-hood (i.e., being verified is either true or false of any ballot manifest in CO), then we encode that with a predicate subtype, `pred[T]`.  This is shorthand notion for a function from `T` to `bool`, `[T -> bool]`.

Comment (kiniry) [inline specs/pvs/corla.pvs]: Thus, @mkiniry is going over the glossary (and eventually the C.R.S. and law) to double-check that each instance of such a binary classifier is encoded in our specification in the same fashion.  We may choose to be more explicit/plain-spoken in our syntax, of course, and can avoid the `pred` notation.  But since this is the standard way of writing things in HOL in PVS, I'm doing it that way by default.

File: specs/pvs/corla.pvs (hunk 13 above)
Comment (kiniry) [inline specs/pvs/corla.pvs]: I'm ok with @nealmcb's recommendation.  I suspect that CDOS has their own terminology.  I had to come up with something since the C.R.S. doesn't actually talk about this in detail.

File: specs/pvs/corla.pvs (hunk 15 above)
Comment (kiniry) [inline specs/pvs/corla.pvs]: Yes, I capture that in the @trace annotation and will write the constraint, encoded in an axiom, later.  Note that `TBD` is simply a synonym for `true` at the moment.

File: specs/pvs/corla.pvs (hunk 16 above)
Comment (kiniry) [inline specs/pvs/corla.pvs]: I've added some exposition to explain such.

File: specs/pvs/corla.pvs (hunk 18 above)
Comment (kiniry) [inline specs/pvs/corla.pvs]: Got it.

File: specs/pvs/corla.pvs [hunk 41]
```diff
@@ -863,6 +934,8 @@ BEGIN
 % @description The audited contest determines the number of ballots
 % that must be examined and verified during the RLA.
```
kiniry: At the moment the information from CDOS is sufficiently underspecified that we have not presume that fractional risk limits are permitted and even desired.  We could, of course, encode non-transcendental risk limits as natural numbers using a mantissa, by why bother?

specs/pvs/corla.pvs
kiniry: Got it.  Thanks.

kiniry: We don't know.  Again, this is mentioned as an adjective in C.R.S., but we do not have clarity on what it means.

specs/pvs/corla.pvs
kiniry: I've refined the datatype to include the public interface.  The interface through which those involved in the audit input ballot interpretations is, conceptually, distinct from the top-level county interface, which is why they are modeled as two different UIs.

specs/pvs/corla.pvs
kiniry: Done.  Good point.

specs/pvs/corla.pvs
kiniry: Yes it is, which is why I have included the text.  We can/should make it clear what parts of the specification are actually realized in the system we are building, but that comes as a part of the description of the refinement, which happens as we write the code this week.

specs/pvs/corla.pvs
kiniry: I'm modeling everything the C.R.S. specifies.  What we implement is a subset of that functionality.  I will add a @todo in the preface to explain how this works and how the reader can understand exactly what is built vs. what is just designed.

specs/pvs/corla.pvs
kiniry: I'll update the text to make it clear that this is currently a quote from C.R.S.  I'll start using LaTeX quotations for such.

kiniry: I've added an axiom for such

specs/pvs/corla.pvs
kiniry: Great, we'll update when we hear back.

specs/pvs/corla.pvs
kiniry: Done.

kiniry: Done, and I have put those paragraphs into a quotation.

kiniry: Thanks for these reflections, @nealmcb.  We'll definitely need to document this in the chapter on RLAs.  I'm leaving the concept in the specification as is at the moment until we get clarify on terminology and use by CDOS.

specs/pvs/corla.pvs
kiniry: Got it; thanks.

specs/pvs/corla.pvs
kiniry: Got it.

kiniry: All good.

kiniry: We must model even actions performed by other people and systems.  They are used to explain the dependencies of our system, or what is also known as its "rely".

specs/pvs/corla.pvs
kiniry: Ok.

specs/pvs/corla.pvs
kiniry: God it.  ;)

specs/pvs/corla.pvs
kiniry: Thanks @nealmcb! As we refine the HOL data model into a concrete one on Monday (I expect), we'll take into account yours for inspiration.

specs/pvs/corla.pvs
kiniry: Good.

specs/pvs/corla.pvs
kiniry: Next time I have you pull out the fine tooth comb, you do it in a branch! ;)

specs/pvs/corla.pvs
kiniry: You are right.  We may flatted these three conceptual tables down into a single one, given that `contest` is a supertype that requires encoding anyway.

specs/pvs/corla.pvs
kiniry: Got it.

specs/pvs/corla.pvs
nealmcb: As I understand it, Dominion doesn't produce a ballot manifest. We've been very clear that the ballot manifest should come from a process completely separate from the scanners. The counties generate it according to our spec as a simple csv file. My csv sample above is based on the CDOS example from the DQ, about which they requested guidance in a place that I don't see offhand.

specs/pvs/corla.pvs
nealmcb: The glossary describes these terms.

docs/glossary_concepts.txt
nealmcb: It is a ballot which cannot be located. The current glossary definition uses it for both counted an uncounted ballots, which we might want to separate out, though the treatment would be the same.

specs/pvs/corla.pvs
nealmcb: I think `wrong_outcome` should be derived from `contest_outcome`, not `election_outcome`, which I don't think we know enough to define.

I agree, we should deferring the issue of multiple kinds of winners until we have CVR data that we know can express it.

docs/glossary_concepts.txt
```diff
@@ -0,0 +1,177 @@
...
+* **political party** TBD
```
nealmcb: Comment on line 103, which is above this section in the file.  Context:
```
%%% The Colorado election law pertaining to election audits is "TBD",
%%% and the rules pertaining to such are found in the "Revised Draft
%%% of Proposed Rules" whose latest version is labeled CCR 1505-1 and
```
There is a distinction between Colorado law (passed by the legislature) and Rules adopted by CDOS. The law on RLAs last changed in 2013 IIRC. So I'd say
```
%%% The Colorado election rules pertaining to election audits is "TBD".
%%% We are working from the "Revised Draft
%%% of Proposed Rules" whose latest version is labeled CCR 1505-1 and

File: specs/pvs/corla.pvs (hunk 18 above)
Comment (nealmcb) [inline specs/pvs/corla.pvs]: This situation should only affect tiny numbers of ballots and seems pretty easy to defend. If we don't assume that, we're diverging from published RLA procedures, and need an alternative in which the math works. I think we should handle it this way and cover it in the documentation for preparing manifests.

File: specs/pvs/corla.pvs (hunk 41 above)
Comment (nealmcb) [inline specs/pvs/corla.pvs]: I think we do want it to be a nonneg_real, for flexibility.

File: specs/pvs/corla.pvs (hunk 19 above)
Comment (nealmcb) [inline specs/pvs/corla.pvs]: Updated text was:
```
%%% interpretations of paper ballot records by a computer. They
%%% frequently, but not always, contain an interpretation of all voter
%%% choicdes in all contests on a ballot. CVRs are sometimes
```
s/choicdes/choices/

File: specs/pvs/corla.pvs [hunk 43]
```diff
@@ -372,8 +416,12 @@ BEGIN
 % @todo Each scanner has a unique id number.
   vvpat: TYPE
   verified?: TYPE = pred[storage_container]
-  log: TYPE
-  chain_of_custody_log: TYPE FROM log
+
+%%% Voting system's logs are precisely defined in federal
+%%% standards. Those standards are known as the ``Voluntary Voting
+%%% System Guidlines'', or VVSG for short.
```
nealmcb: I've been involved in the VVSG update discussions on event logging, and they are good on syntax, but near worthless on semantics. The adopted VVSG standards do give some precise definitions, but I don't know if they cover anything relevant for us.

docs/glossary_concepts.txt
```diff
@@ -0,0 +1,177 @@
...
+* **undervote** TBD
+* **uploading interface** A part of the county dashboard.
+* **user experience (UX)** TBD
+* **user interface (UI)** TBD
+* **username** TBD
+* **voter marking** TBD
+* **voting method** TBD
+* **voting system** TBD
+* **working group** TBD
+* **wrong outcome** - When the reported outcome for a given contest does not match the
```
kiniry: Thanks for those reflections.  We'll go ahead and add them to the glossary and specification, realizing that some may end up being synonyms and others may end up being eliminated, just like all other concepts introduced and evaluated.

kiniry: Added in cd5e86523037b5f60e11c91bd06854fedbfdcb9b.

docs/glossary_concepts.txt
kiniry: But I actually wanted to cite the law (as passed by the legislature) in the TBD.

specs/pvs/corla.pvs
kiniry: Agreed.  Perhaps I shouldn't have used the word "precise". ;)

specs/pvs/corla.pvs
dmzimmerman: Agreed.

dmzimmerman:
I believe this is now cleanly merged and ready to actually be merged back into master; however, since I resolved the merge conflicts, I probably ought not to be the one who actually merges it into master just in case I did something boneheaded.

docs/glossary_concepts.txt
nealmcb: LOL. I took it as a subtle note that we're working from an unfinished spec....

The law is at Section 1-7-515, C.R.S.  The official online site is lexisnexis, which includes a handy "history" section, but unfortunately they don't provide linkable resources.  This is a linkable version from Justia: http://law.justia.com/codes/colorado/2016/title-1/general-primary-recall-and-congressional-vacancy-elections/article-7/part-5/section-1-7-515/

specs/pvs/corla.pvs
```diff
@@ -6,7 +6,18 @@
 %%%
```
nealmcb: At line 866, I would change this comment for `  margin_overstatement: TYPE = nat`:
```
% @description A circumstance in which the audit board’s
% interpretation of ballot markings reveals that the winner with the
% fewest votes received fewer votes than the voting system’s
% interpretation of the same markings as reflected in the CVR.

% @example For example, if the CVR reflects an undervote in the
% audited contest, and the audit board’s interpretation of the
% corresponding paper ballot reflects a vote for the loser with the
% most votes in that contest, the CVR contains a one-vote
% overstatement. if the CVR reflects a vote for the winner with the
% fewest votes, and the audit board’s interpretation of the paper
% ballot reflects a vote for the loser with the most votes, the
% preliminary results contain a two-vote overstatement.
  margin_overstatement: TYPE = nat
```
to....

```
% @description A circumstance in which the audit board’s
% interpretation of ballot markings reveals that the winner with the
% fewest votes had a smaller margin than in the voting system’s
% interpretation of the same markings as reflected in the CVR.
```
and capitalize the "if" that starts the second example in the next paragraph.

A subtle difference, but just to be clear.

It needs to change in a similar way for the understatement definition.

kiniry: Done.

kiniry:
I have resolved all comments and conflicts, so we are merging this back into master so that we can branch again to work on protocol and polish.

PR #86, opened 2017-07-21
Smart CI

ranweiler:


nealmcb:
Offhand, looks good.
I suggest adding some documentation to note that if someone adds a global config file or some other way that client or server tests depend on outside files, this optimization might miss something.
What tests have you run?

ranweiler:
@nealmcb previously, this branch had commits that changed the READMEs in the relevant subdirectories. I then killed them and force-pushed, one-by-one, to make sure clean directories wouldn't have their checks spuriously run.

ranweiler:
I'll add a comment about that edge case. @kiniry, would you like to see that "gotcha" flagged anywhere special?

nealmcb:
Re the gotcha about how building is optimized, I'd suggest noting it in docs/25_developer.md

ranweiler:
@nealmcb good call! I'll add something there, as well as a comment in the CI file itself.

nealmcb:
Those are great warnings, ranweiler. I'm good with this now, but I guess I can only recommend merging.

dmzimmerman:
Also, given that we actually have documentation and specifications that build, we should probably add such builds (typechecking of the PVS, generation of the PDF from the PVS, generation of the PDF from the Markdown) to continuous integration at some point. A topic for another PR, when we have more time.

PR #95, opened 2017-07-22
Glossary specification double check

mkiniry:
I'm working with Joe to create a pull request so I can begin learning this process.

### DIFF
--- a/docs/89_glossary.md
+++ b/docs/89_glossary.md
@@ -507,3 +507,14 @@ reported tabulation outcome.
 
 * **Slack** TBD
 
+* **Paused Audit** @review morganmillerux I'm unclear if they will be allowed to pause and audit but it seems valuable to state what it is and then explicitly allow or disallow it.
+
+* **Missing/Not Found Ballot** @review morganmillerux this is perhaps the same as 'phantom ballot'
+
+* **Credentialing Authority** @review morganmillerux Do we need a concept of who's allowed to author credentials?
+
+* **Submit Ballot Interpretation** @review morganmillerux I see an entry for Audit Results but I think it would also be helpful to define the granular steps of submitting each ballot interpretation.
+
+* **Ballot Interpretation** @review morganmillerux Perhaps I didn't a synonym for this, but this should be included.
+
+* **Ballot Retrieval** @review morganmillerux the process of getting the ballots seems like a specific point of concern and merits being called out.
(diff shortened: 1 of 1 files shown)


docs/20_tutorial.md
```diff
@@ -2,3 +2,63 @@
...
+
+***should the tutorial deal with the SOS establishing the risk
+limit, or any of their pre-audit responsibilities like the selection
+of which races to audit?
+
+Brief introduction to RLAs in non-technical, human friendly language.
+At the county level, something along the lines of "This tool is going
+to help you cross reference votes recorded on paper ballots with how 
+the computers read those ballots, to determine whether the computers
+did their job accurately…” Try to put their minds at ease about the
```
nealmcb: Agreed.  But I would suggest substituting "system" for "computer", since the CVR represents the end result of a complex interaction of equipment, computers, and humans doing manual adjudication on some questionable ballots. Perhaps:
> How the system interpreted those ballots...
> to determine whether the system functioned properly

docs/20_tutorial.md
```diff
@@ -2,3 +2,63 @@
...
+process. :)
+
+***should the tutorial cover the mechanics of storing ballots and the
+other physical aspects of the RLA process which are not ‘in’ the RLA
+tool/system?
```
nealmcb: I think not the mechanics, which vary by voting system, and even by county.
But we'd love to refer to best practices developed by counties and CDOS.

docs/89_glossary.md
```diff
@@ -402,7 +504,7 @@ reported tabulation outcome.
 
 * **county number** as part of the ballot manifest
```
nealmcb: We seem to be using county names in the manifest, following the original CDOS example.

nealmcb: The definitions for "ballot manifest" and "cast vote record" should speak in terms of "ballot cards", rather than "ballots", they now do.
A ballot may be made up of multiple cards. A "ballot card" is a single sheet of paper, which may have contests on both sides. Each ballot card is tabulated independently, and reflected in the CVR report as a separate row. The "ballot style" for all cards of a ballot is either always or often the same for all of them, from the standpoint of the voter, though the software will consider the style to be the actual set of contests on the ballot card.

kiniry: I have updated the full glossary wrt ballots vs. ballot cards.

PR #101, opened 2017-07-23
Cherry-pick of single important commit in borked mock-up-review branch.

kiniry:
Please review for my cherry-picked single commit of relevance.  I intend to delete the old mock-up-review branch after we merge this.

### DIFF
--- a/specs/pvs/corla.pvs
+++ b/specs/pvs/corla.pvs
@@ -248,14 +248,22 @@ elections: THEORY
 BEGIN
 %%% Voters are the most important concept in elections, as they are
 %%% who elections are for, and who make the choices that determine the
-%%% outcome of legitimate elections. Some voters are UOCAVA
-%%% voters. What follows is one way to formalize such, simply stating
-%%% that the type uocava\_voter is a predicate on the type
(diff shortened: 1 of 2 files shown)


specs/pvs/corla.pvs
```diff
@@ -248,14 +248,22 @@ elections: THEORY
...
+    [# firstname: string,
+       lastname: string,
+       party_affiliation: political_party #]
   elector: TYPE = voter
-  uocava_voter: TYPE from voter
+%%% Some voters are UOCAVA voters. What follows is one way to
+%%% formalize such, simply stating that the type
+%%% \texttt{uocava\_voter} is a predicate on the type
+%%% voter. Consequently, all voters are either UOCAVA or not.
+  uocava_voter: pred[voter]
```
nealmcb: I know this commit is merged, so ignore this comment if desired.  But I'm unclear on the value of the term uocava_voter, or the modeling of a firstname and lastname but nothing else. I don't see the term UOCAVA in the draft rule, and non-uocava ballots can come from uocava voters. So I'm unclear on the benefit of these sections. I think from the auditing perspective, non_voter_verifiable_ballot describes what we need to deal with, and even that maybe not this go round.

PR #104, opened 2017-07-24
Data model prose

sfsinger19103:
Trying to make the document more accessible to CDOS and other readers


specs/pvs/corla.pvs
```diff
@@ -1,33 +1,78 @@
...
+%%% case of Colorado, it facilitates running risk limiting audits
+%%% across all counties in the state simultaneously.
+%%%
+%%% The Colorado election law pertaining to election audits is
+%%% \href{http://law.justia.com/codes/colorado/2016/title-1/general-primary-recall-and-congressional-vacancy-elections/article-7/part-5/section-1-7-515/}
+%%% {Section 1-7-515, C.R.S.}  and the rules pertaining to such are
+%%% found in the "Revised Draft of Proposed Rules" whose latest
+%%% version is labeled CCR 1505-1 and dated July 6, 2017. Section 25,
+%%% Post-Election Audit, of that document is the salient portion of
+%%% that document. We will call this document RDPR-6-Jul-2017 for the
```
nealmcb: that document _for our purposes_

specs/pvs/corla.pvs
```diff
@@ -40,16 +85,14 @@
 %%% informal specifications which will be extracted using a shell script
 %%% into a well-typed FAFESSL informal domain model specification.
 %%%
-%%% As mentioned above, we write formal specifications using a
-%%% literate style, a la Knuth. Doing so permits us to document our
```
nealmcb: "literate" and "literate programming" are technical terms, but don't really sound like it. Thus I think it makes sense to retain the reference to Knuth, or perhaps, better, something like https://en.wikipedia.org/wiki/Literate_programming, to help clarify for readers what is going on.

dmzimmerman: I agree. The reference could be moved - perhaps to the first place where "literate" is mentioned - but should stay somewhere.

specs/pvs/corla.pvs
```diff
@@ -1,33 +1,78 @@
...
-%%% of the system delivered to the Colorado Department of State. The prose
-%%% annotations should help Colorado Department of State elections domain experts
+%%% of the elements of the specification. 
+%%% The specification documents our
+%%% thinking and the system design, from domain modeling and
+%%% engineering all of the way to formal specification and
+%%% verification.
+%%%
+%%% We write the formal specification using a
+%%% literate style, allowing the production of typeset books
```
dmzimmerman: This would be a good place for the reference to Knuth and literate programming.

specs/pvs/corla.pvs
```diff
@@ -1,33 +1,78 @@
...
+%%% and interactive hypertext directly from the specification.
+%%% The prose
+%%% annotations will help Colorado Department of State elections domain experts
 %%% and technical staff to assess the completeness and correctness of the
-%%% ColoradoRLA system/tool.
+%%% ColoradoRLA tool.
```
dmzimmerman: Is it the RLA Tool or the ColoradoRLA Tool?

specs/pvs/corla.pvs
```diff
@@ -1,33 +1,78 @@
...
+%%% in Java and runs on servers hosted by CDOS. The client is
+%%% implemented in TypeScript and JavaScript and runs in modern web
+%%% browsers.
+%%% CDOS provides
+%%% the hosting servers, network, and a number of services
+%%% in support of the tools deployment and management.
 %%%
 %%% This document is written in a literate style using the PVS
 %%% theorem prover. Writing system specifications in this fashion is
 %%% called (formal) domain engineering. Our formal domain model is
```
dmzimmerman: If we're removing "our" in the line below, we should remove it here too.

specs/pvs/corla.pvs
```diff
@@ -203,9 +209,13 @@ END information_systems
 %%% requirements. Preferably redundant servers would be hosted in
 %%% separate facilities, with separate and independent power systems
 %%% and networks.
-%%%
```
dmzimmerman: Are you intending to merge two paragraphs here? (that's what's happening with the LaTeX document when you delete this line)

specs/pvs/corla.pvs
```diff
@@ -203,9 +209,13 @@ END information_systems
...
 %%% An Election Night Reporting (ENR) server is mentioned repeatedly
 %%% in the RDPR-6-Jul-2017, so we model it here.
+%%% The next few lines of the specification encode the process of establishing
```
dmzimmerman: Similarly, is this (line 214) now meant to be a paragraph on its own?

specs/pvs/corla.pvs
```diff
@@ -269,14 +276,16 @@ BEGIN
...
   candidate: TYPE FROM person
 
-%%% Elections focus on contests, each of which represents a set of
-%%% choices that voters can make. Making a legal choice ranges
+%%% Elections focus on contests, each of which represents a 
+%%% choice that voters can make. Making a legal choice ranges
 %%% enormously across the Earth, from marking a vote for a candidate
 %%% by filling in a single bubble to enumerating a total order on all
-%%% choices in a contest. Across a given ballot, each contest has a
+%%% choices in a contest. Within any single elction, each contest has a
```
dmzimmerman: election

specs/pvs/corla.pvs
```diff
@@ -1,33 +1,78 @@
...
+%%% the hosting servers, network, and a number of services
+%%% in support of the tools deployment and management.
 %%%
 %%% This document is written in a literate style using the PVS
 %%% theorem prover. Writing system specifications in this fashion is
 %%% called (formal) domain engineering. Our formal domain model is
-%%% written in PVS's higher-order logic (HOL). Our informal domain
-%%% model is written in the Free \& Fair System Specification Language,
+%%% written in PVS's higher-order logic (HOL). There is a corresponding informal domain
+%%% model written in the Free \& Fair System Specification Language,
```
nealmcb: Isn't that Free & Fair*_Election_ System Specification Language, here and a few other places? (quoting https://github.com/FreeAndFair/BON)?  Or are you re-thinking the name and positioning it as more general? Though the "E" in FAFSSL is handy for pronunciation. It could be "Free & Fair Extended System Specification Language I guess.

dmzimmerman: Well, it really _is_ more general, given that the specification language itself has nothing to do with election systems (heck, depending on where the funding for actually building the tooling for it comes from, it might be GSSL - the Galois System Specification Language - which is how we've referred to it the past few years). I'd actually drop the A from the acronym and make it "FFSSL", if I had to choose something this minute.

dmzimmerman: (or "KZSSL" :P )

nealmcb: Beware the "Freakishly Fast and Simple Scripting Language" project.
KZSSL is more fun to pronounce, and doesn't seem to conflict with anything ;)

specs/pvs/corla.pvs
```diff
@@ -69,12 +112,16 @@
...
 
-%%% Our formal domain model is also annotated with structured comments
-%%% in precise natural language using a standard set of annotations.
-%%% A shell script processes these annotations and generates a
-%%% well-typed informal FAFESSL specification.
+%+BEGIN background
+
+%%% \section{Formal Model Description}
+
+%%%\subsection{Basic Technical Infrastructure}
```
nealmcb: Missing space after %%%, here and one other place.

specs/pvs/corla.pvs
```diff
@@ -92,35 +139,9 @@
...
-
-% @todo kiniry Add a discussion of our glossary and its relationship
-% to this specification.
-
-% @todo kiniry Define C.R.S. and explain the other sources used to
-% derive all concepts in this specification.
-
 % @todo Explain how refinement works and how the reader can understand
 % exactly what is built vs. what is just designed.
+% @review kiniry what's the point of including these annotations? I don't see them used in the document @sfsinger
```
nealmcb: Which annotations do you mean? The @todo above, or the Literate PVS notes below? Both seem used and useful to me.

specs/pvs/corla.pvs
```diff
@@ -136,23 +157,8 @@ END fafessl
...
-%%% in Java and runs on servers hosted by CDOS. The client is
-%%% implemented in TypeScript and JavaScript and runs in modern web
-%%% browsers.
-%%%
-%%% As we are hosting the RLA Tool in CDOS's system, they are
-%%% providing the hosting servers, network, and a number of services
-%%% in support of the tools deployment and management.
-%%%
-%%% A database is used to store all of the data relevant to
+%%%% A database is used to store all of the data relevant to
```
nealmcb: Extra %

specs/pvs/corla.pvs
```diff
@@ -269,14 +276,16 @@ BEGIN
...
 %%% enormously across the Earth, from marking a vote for a candidate
 %%% by filling in a single bubble to enumerating a total order on all
-%%% choices in a contest. Across a given ballot, each contest has a
+%%% choices in a contest. Within any single elction, each contest has a
 %%% unique name.
 % @todo kiniry Need to break the cycle between the contest type and
-% the ballots theory.
+% the ballots theory. sfsinger suggests: contests are grouped into elecitons,
+% not ballots. So define contests first. Then a ballot is created to 
+% represent a set of contests.
```
nealmcb: The work of the NIST/EAC Interoperability Working group, Election Modeling subgroup, might be useful. (Yeah, pretty late to bring this up, but I just noticed their recent progress, FWIW).  The seem to have a BPMN2 model, pictures, XPDL, xml, who-knows-what at https://github.com/usnistgov/ElectionModeling.  See also
http://collaborate.nist.gov/voting/bin/view/Voting/ElectionModeling
Also the UML for the https://github.com/usnistgov/CastVoteRecords work.

specs/pvs/corla.pvs
```diff
@@ -394,24 +409,37 @@ END election
...
+% as subsets of "contests". A ballot question is a contest between "yes" and "no".
 %%%
-%%% Ballots are stored by one of several possible means. Some ballot
+%%% There are a variety of ways to store ballots. Some ballot
 %%% storage containers are secure, others are not. Some facilitate an
 %%% easy means by which to find a particular ballots, others do not.
 
 % @relation box inherit storage_container
 % @relation bin inherit storage_container
 % @relation ballot_box inherit storage_container
```
nealmcb: Is there supposed to be a connection between these notions of boxes vs bins and the notions of efficiency or secure vs insecure storage? Why are we distinguishing these types of storage?

specs/pvs/corla.pvs
```diff
@@ -625,11 +667,11 @@ END ballots_collections
 
 %+BEGIN ifr
 
-%%% \section{Instructions, Forms, and Reports}
+%%% \subsection{Instructions, Forms, and Reports}
 
 %%% Elections have all kinds of different published instructions,
 %%% forms, and reports. We enumerate a few here that are specific to
-%%% Colorado, and the RLA Tool in particular.
+%%% Colorado, and relevant the RLA Tool in particular.
```
nealmcb: relevant _to_

specs/pvs/corla.pvs
```diff
@@ -758,20 +802,23 @@ BEGIN
...
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-%%% \section{Cryptography}
+%%% \subsection{Cryptography}
 
-%%% We will add additional cryptographic primitives here as required.
+% We will add additional cryptographic primitives here as required.
+% @review kiniry dmz Is it correct to say that the cryptography is used
+% (via hashing) to verify that the ballot manifests and cvrs stored in the 
+% database are actually the ones uploaded by the Counties?
```
nealmcb: The hash confirms first of all that the file upload proceeds without any corruption, when it is checked after receipt.  It can also be used by counties, CDOS and the public to confirm that everyone is talking about exactly the same file, after it is downloaded again. It looks like they will be stored as rather large binary objects in Postgres. I hope it and our deployment / design can handle objects that are pushing a gigabyte in size, though we won't see that in this state, this year.

specs/pvs/corla.pvs
```diff
@@ -781,17 +828,18 @@ END cryptography
 
 %%% \subsection{Randomness}
 
-%%% High quality randomness is key to the legitimacy of any
-%%% risk-limiting audit.
+%%% Choosing a random sample of ballots is key to the legitimacy of any
+%%% risk-limiting audit. Legitimacy depends on a high-quality process for
+%%% the random samples. The Colorado RLA Tool will produce a random order 
+%%% of all the ballots once and for all; then random samples can be created
```
nealmcb: => of all the ballots _for each county_ once and for all; then random samples _of the necessary size_ can be created

specs/pvs/corla.pvs
```diff
@@ -1001,23 +1050,29 @@ END rlas
...
 %%% syntactically written as comma-separated values, and other times
 %%% in plain English. Some CVRs use a election-specific encoding
 %%% scheme to represent choices (e.g., a '1' means "John Doe"); others
 %%% use plain English.
 %%%
+%%% CVRs may (and in the case of CVRs exported from Colorado's Dominion system, do)
+%%% contain information about the ballot beyond the voter choices. For 
+%%% example, ballot style, precinct id and ballot type (in-person vs. mail)
+%%% may be included.
+%%%
```
nealmcb: `CountingGroup` is the one that are "Mail" or "In Person". `BallotType` is the ballot style, which includes a precinct number which is a big problem from a CORA standpoint. We would prefer that all counties upload a CVR which has all the precinct information left out: `PrecinctPortion`, the precinct part of `BallotType`, and `ImprintedId`.

specs/pvs/corla.pvs
```diff
@@ -1001,23 +1050,29 @@ END rlas
...
 %%% CDOS requirements mandate that CVRs are exported and uploaded to
 %%% the RLA back-end as CSV files by county officials using their
 %%% voting systems and the RLA Tool, respectively.
 
 % @todo kiniry For a given election, there must be a 1-1
 % correspondence between its ballots and its CVRs.
+% @review kiniry between its ballots and the data-rows of its CVR(s), right?
```
nealmcb: I don't think we can actually always expect either a 1-1 correspondence between paper ballots and CVRs (given phantom ballots) or manifest entries and either of those (human error, to be handled also via phantom ballots).

specs/pvs/corla.pvs
```diff
@@ -1043,14 +1098,18 @@ END cast_vote_records
...
 %%% The user interfaces of the system are the visible interactive
 %%% parts of the application. There are three different user
 %%% interfaces in the RLA Tool. The precise names of these interfaces
 %%% are still under discussion; one is for CDOS personnel responsible
 %%% for the audit at the state-level, one is for county personnel at
 %%% the county-level, and one is for audit board members.
+%%% Note that all these user interfaces both receive and 
+%%% provide information.
+%%% Because the public dashboard only pushes information out, 
+%%% we do not consider it a "user interface".
```
nealmcb: I think we ought to add the public dashboard to the list.  It public dashboard will not change the database, but will accept some sorts of interaction and/or input, whether mouse clicks or searches or the like.

specs/pvs/corla.pvs
```diff
@@ -1283,8 +1348,6 @@ BEGIN
 
 % @trace RDPR-6-Jul-2017 Section 25.2.2(K)
 
-%%% \subsection{Random Selection of Ballots for Audit}
```
nealmcb: Did you mean to move this up above this line?
`%%% The state-wide dashboard will allow the Secretary of State`

specs/pvs/corla.pvs
sfsinger19103: I'll leave action on these comments to others.

specs/pvs/corla.pvs
sfsinger19103: One random *order* of all ballots state-wide allows random samples restricted to counties: for a random sample of size n for County X, simply take the first n ballots from County X that appear in the statewide random order.

nealmcb: That would work, but I'm thinking it would make it harder for the public to verify the selection, since now they have to include all manifests when they do so, in some canonical order among counties, and then do their own winnowing of the overall list to generate what is selected for a given county. They couldn't e.g. just feed a single county manifest (with a bit of tweaking) into Philip's tool and confirm the sample that way. I'll have to think about this....

specs/pvs/corla.pvs
nealmcb: It does seem better with a paragraph break here, and for the comment above.

kiniry:
I will review and work on this as soon as PR #108 is merged.

kiniry:
PR #108 is now merged, so I'll look at this now.

specs/pvs/corla.pvs
kiniry: I was distinguishing them because (a) I had witnessed different storage mechanisms in the field, (b) I understand there is not a uniform mechanism in CO, and (c) some mechanisms are secure and provide a chain-of-custody evidence and others are not.  This latter point matters wrt the meaning of `secure` and `verified`.

nealmcb:
Some comments from my continuing review of what was merged:

change 'undervote' to 'blank vote' most everwhere

at some point fix comment around audited_ballot_by_contest_list_table
 https://github.com/FreeAndFair/ColoradoRLA/pull/108/commits/24ab0327e1b82a2737963a7adc98496b70ad3193#r130158627

> +%%% This domain model is based upon these statues, its associated

=> statutes

> +%%% traceability relation, then the reason for its introduce should be

=> introduction

> %%% apply a hash value to the CVR export file using the hash value

should we confirm this hash? need to know what it is - sha256?

specs/pvs/corla.pvs
kiniry: Making sure that our model is consistent with federal standards someday is a good thing, but we don't have time or resources for it now. What consistent usually means is that there is a relation we can define between types in the models that is information preserving from the simpler model to the more complex one.

specs/pvs/corla.pvs
kiniry: I'm in alignment with everything that @nealmcb stated here.  Thinking carefully about foolproof means by which 100s of megabyte uploads can happen reliably and securely is important.  Historically I see government agencies purchasing commercial products for such services but they don't work as well as the off-the-shelf open source stuff we use.  E.g., I would think that a secure rsync, or at least a restartable sftp, is wiser than anything else in this regards.

specs/pvs/corla.pvs
kiniry: Understood. In a perfect world, they would be 1-1, but that'll basically never happen.  And it seems like different vendors do different things about the relationship between CVRs and multi-page ballots.

specs/pvs/corla.pvs
kiniry: I agree with you @nealmcb.  A UI is a UI, regardless of if it is read-only or not. We need to specify events and scenarios even for the public dashboard, in this regards.

specs/pvs/corla.pvs
nealmcb: A first easy step is to support automatic compression. They give us a csv, we compress it (as a .zip?), upload it, process it (ideally on-the-fly) and store it. Then offer the compressed version for download.

nealmcb: I think CDOS does have some sort of secure ftp infrastructure for counties, which we might be able to leverage, by letting them point to a reference to the uploaded file which we could pull from.

PR #108, opened 2017-07-24
RLA Tool protocol specification

kiniry:
The RLA Tool protocol is specified in the formal system specification (`corla.pvs`).  Its root is each function declaration that operates on a subtype of `dashboard`.  The types used in those signatures denote which types must be serialized in JSON.  The types that are mentioned in the data model—a superset of types serialized in the protocol—must be serialized to the DB.  A sequence diagram will summarize all dashboard message exchanges in a fashion that the client will immediately grok.


specs/pvs/corla.pvs
```diff
@@ -1193,10 +1205,18 @@ BEGIN
...
 %%% five percent.
-  establish_risk_limit_comparison_audits:
-    [department_of_state_dashboard, {r: risk_limit | r <= 5} -> department_of_state_dashboard]
-  establish_risk_limit_ballot_polling_audits:
-    [department_of_state_dashboard, {r: risk_limit | r <= 5} -> department_of_state_dashboard]
+  establish_risk_limit_for_comparison_audits:
+    [department_of_state_dashboard, {r: risk_limit | r <= 5} ->
+     department_of_state_dashboard]
+  establish_risk_limit_for_ballot_polling_audits:
+    [department_of_state_dashboard, {r: risk_limit | r <= 5} ->
```
nealmcb: Since CDOS retains the right to change anything in the real audit, I'd leave out an enforced check on the risk limit being less than 5%. Both counties and CORLA2 have supported a move to 10% for county-level contests, and they may want flexibility.

specs/pvs/corla.pvs
```diff
@@ -2022,7 +2066,111 @@ BEGIN
...
+%%%
+%%% As discussed in~\autoref{sec:Department_of_State_Dashboard}, the
+%%% Department of State Dashboard has eight (8) features: one for
+%%% authentication, six for providing information about a given
+%%% election and RLA, and one for getting updates on the current RLA.
+%%%
+
+%%% \begin{enumerate}
+
+%%% \item \texttt(authenticate\_state\_administrator): authentication
```
ranweiler: Is this distinct from authenticating as a user more generally (i.e. "logging in ")?

specs/pvs/corla.pvs
nealmcb:
```
%%% @trace RDPR-6-Jul-2017 Section 25.2.4
%%% FOR THE 2017 COORDINATED ELECTION, THE SECRETARY OF STATE MAY, BY ORDER,
%%% ALTER ANY OF THE REQUIREMENTS OUTLINED IN RULE 25.2.
```

specs/pvs/corla.pvs
```diff
@@ -2022,7 +2066,111 @@ BEGIN
...
+%%% server(s) is initiated by the client, as with normal REST-based
+%%% architectures. As such, functions defined in the dashboards
+%%% (\autoref{sec:Dashboards}) of the system are the endpoint of all
+%%% communication in the system. The RLA Tool protocol is consequently
+%%% the union of all of those endpoints.
+%%%
+%%% \subsection{State-wide Dashboard Protocol}
+%%% \label{sec:State-wide_Dashboard_Protocol}
+%%%
+%%% As discussed in~\autoref{sec:Department_of_State_Dashboard}, the
```
ranweiler: Will the items in this section correspond to exactly one URL endpoint each, or potentially several? Right now, what this describes seems to be more RPC over HTTPS. To be clear: that's totally fine if we want to do that.

But, since the above spec talks about REST, I'm wondering if we plan on further specifying how to realize these actions via HTTP methods applied to URL-addressable resources.

ranweiler: Also: if this PR is not meant to address that level of implementation specificity, feel free to let me know.

specs/pvs/corla.pvs
dmzimmerman: I think we need to make a choice about how that's going to work - are there going to be separate endpoints for authentication to the different roles, or one endpoint that will result in an authentication to one of a number of roles (whichever the user is authorized for).

specs/pvs/corla.pvs
dmzimmerman: As I read it, each item corresponds to one URL endpoint. They clearly need more specifics with respect to what information has to be supplied in the request for each and what will be provided in the response. I'm not sure about "refresh"; it might make sense for there to instead be individual endpoints for you to use to retrieve the pieces of information you're displaying on various parts of the dashboard.

specs/pvs/corla.pvs
ranweiler: @dmzimmerman, from my perspective, either is fine. IMO, you should select whichever strategy is easiest to implement correctly. For phase 1 at least, I'm organizing the client pages around the idea that a user is logged in as an SoS xor County Official, though the design allows for a user to be both if that ends up being desired. To make implementation efficient, I am explicitly not assuming any sort of granular role-based access control, but that authorization will happen at the granularity of the two user types.

specs/pvs/corla.pvs
ranweiler: > I'm not sure about "refresh"

As long as I know what URLs contain data for different browser client pages, we can poll the server. For phase 1, we _could_ have a very coarse "tell me about the whole world" endpoint, and break it up for phase 2+.

specs/pvs/corla.pvs
dmzimmerman: In that case, let's have two authentication endpoints, one for each role, to correspond to the two authentication functions already defined in the "authentication" theory.

specs/pvs/corla.pvs
```diff
@@ -2022,7 +2066,111 @@ BEGIN
...
+%%% the elections under audit,
+
+%%% \item \texttt{publish\_seed}: the publication of the random seed
+%%% for RLAs in the election under audit,
+
+%%% \item \texttt{establish\_risk\_limit\_for\_comparison\_audits}:
+%%% establishments of risk limits for the election under audit,
+
+%%% \item \texttt{select\_contest\_for\_comparison\_audit}: selection
+%%% of audited contests for the election under audit,
```
nealmcb: offhand I don't see any reason for the `\_for\_comparison\_audits` text in three of these (in the previous 11 lines), since risk limits mean the same thing for both comparison and other risk_limiting audits.

specs/pvs/corla.pvs
```diff
@@ -2022,7 +2066,111 @@ BEGIN
...
+%%%
+
+%%% \begin{enumerate}
+
+%%% \item \texttt{authenticate\_state\_administrator}: authentication
+%%% of state administrators (the mandatory first step in the
+%%% protocol),
+
+%%% \item \texttt{select\_contest\_for\_comparison\_audit}: defining
+%%% the elections under audit,
```
nealmcb: Is this meant to be `select\_election...` rather than `select\_contest`?

specs/pvs/corla.pvs
```diff
@@ -1947,12 +1989,12 @@ BEGIN
   authentication_table: TYPE =
     sequence[[administrator, two_factor_authentication]]
 
-%%% From the State's point of view, the State-wide dashboard permits
-%%% State administrators to establish risk limits for each contest,
-%%% specify which kinds of audits are used for which races and in
-%%% which counties, etc.  Most of the dynamic data available to State
-%%% administrators is provided via County administrators using the
-%%% County dashboard.
+%%% From the State's point of view, the Department of State dashboard
```
nealmcb: Drop "From the State's point of view,"

specs/pvs/corla.pvs
```diff
@@ -132,14 +132,21 @@
 fafessl: THEORY
```
nealmcb: Attached to this top line I'll enter a number of comments on the data model proper, to keep them all together.  Sorry I didn't get to them earlier.
Line numbers refer to commit 5abd84e or 140519f

nealmcb: Line 307:
```
  election_outcome: TYPE = set[contest_outcome]
```
I would change that to `election_outcomes`, since it is not a summary, 
and the contest outcomes are more or less independent.

nealmcb: Line 320
```
  rla_tabulation_outcome: TYPE FROM election_outcome
```
I'm not sure what this is for.  Is it a corrected `reported_tabulation_outcome` based on audit discrepancies?

nealmcb: Line 570:
```
ballot_manifest: TYPE = set[ballot_manifest_info]
```
Should be a sequence? They want to be able to sort selections in order by the manifest, i.e. by "ascending order of Scanner/Batch/Position nos." (Audit Board Reporting Utility Mock-up.googledoc),

I note the database already has them in order in `ballot_manifest_table`, if that order is actually the same as Scanner/Batch/Position nos.

nealmcb: Line 1036:
```
  cvr: TYPE = set[vote]
```
add in `ballot_manifest_info`
We want that info for `verified?`, `compare_reported_markings_to_cvr`, `cvr_table`.
The only related change would seem to be to drop `ballot_manifest_info` from  `ballots_under_audit_table`, since it will already be in `cvr`

nealmcb: Line 1039:
what is `as_csv` for?  Export a single cvr as a csv file?

nealmcb: Line 1980:
```
  national_contest_table: TYPE =
    sequence[[nation, contest]]
  state_contest_table: TYPE =
    sequence[[state, contest]]
```
I don't see a need for these. I'd think just the county_contest_table was sufficient.
At some point for cross-county contests, we'll need to rely on the contest name being uniform across the counties.

nealmcb: Line 1989:
```
  ballots_to_audit_list_table: TYPE =
    sequence[[contest, sequence[ballot_manifest_info]]]
```
We should select ballots by county, not by contest. I'm not sure why we would store it this way.
The county selects from one list of ballots to audit, and audits all the contests on it.
When multiple contests are being audited, we wouldn't want disjoint sets of selections.
When a county only appears on a subset of contests, at this point we just present them all
and work until they audit enough of them.  When, in some future version, we want to give them a filtered list by contest, we may want to filter by multiple (still active) contests, not just one.
Cross-county contests are audited via a yet-to-be solidified stratified audit procedure.

nealmcb: Line 2029:
```
  cvr_table: TYPE =
    sequence[[county, cvr]]
```
We presumably want to index this by ballot_manifest_info, so perhaps that should actually be managed
separately from `cvr`, or just added separately in here if that helps with the indexing (in which case
we can drop `county`.

nealmcb: Line 2036: 
```
  contest_cvr_table: TYPE =
    sequence[[contest, cvr]]

```
I don't see a purpose for organizing or storing things this way across counties.  All cross-county random selections (when we get to them) will be stratified by county, so that they don't end up having disjoint sets of random selections in a single county for multiple contests.

nealmcb: Line 2043
```
  tabulation_result_table: TYPE =
    sequence[[contest, rla_tabulation_outcome]]
```
I see no need to read the county tabulation results file. The file sample they just sent is a real mess.
We should calculate the winners and `contest_margin` values from the already-parsed CVR files.
In terms of checking integrity, the rla_tabulation_outcome is an intermediate result that they didn't include
in the DQ - no sample, no requirement around checking it etc. IIRC.

At least at this point, I think we should probably present the sets of winners that we calculate when presenting
the contests to the audit board for entry, and later in the generated reports and downloads.

nealmcb: Line 2095
```
  ballots_under_audit_table: TYPE =
    sequence[[ballot, ballot_manifest_info, list[ballot_mark],
              cvr, margin_overstatement, margin_understatement]]
```
It's not clear to me how the `ballot` helps here.
It seems that, at least on this go-round, we won't get a list of ballot_marks, but rather in effect a CVR generated by the auditors, which CORLA2 has called an aCVR (audit CVR).
If you accept my suggestion that cvr include ballot_manifest_info, it wouldn't be necessary here.

Having data on when ballot interpretations are added is critical for the sort of efficiency / time-and-motion studies that we'll want to do to learn from this unique experience.  If that data isn't available in some other place, it would be great to add a timestamp for when the audit interpretation was first entered.
Those timestamps should be included in the the data dump that is made available to the public.
Ideally (someday) we would add another one for when it was last modified, if it can be modified.

nealmcb: Line 2101:
```
  rla_summary_results_table: TYPE =
    sequence[[contest, summary_results_report]]
```
It's not clear to me that we need or use this.

dmzimmerman: Seems reasonable for this iteration of the system.

dmzimmerman: I think there may be two different definitions of the word `audit` here. There is `audit` in the sense of "the SOS selected this contest to audit", and then `audit` in the sense of "this ballot is one of the ones selected to be audited as part of the random sequence". So when the county pulls a ballot they enter their interpretations for all the contests (thus auditing the entire ballot), but to perform the RLA calculations for a specific contest, don't we select the random sequence based on the set of ballots on which that contest appears rather than the set of all ballots?

dmzimmerman: Is it "kosher" for us to re-tabulate the entire election ourselves? (it's also unclear where exactly we'd do that; we don't have a database table with individual columns for the individual fields of a CVR - eliminating the easy "ask the database for the sum of these columns" method of counting - so it'd have to be done by the Java code in memory somewhere and stored... where?)

nealmcb: It is critical for someone to confirm that the CVR data we're auditing has the same outcome as the reported outcome, and that the margins reflected in it match the margins we're using for the sample size calculations. Otherwise, they could send us CVRs that match the ballots but not the announced results, or set the margin at 50%, or all kinds of things. The natural thing is for the tool to do that, if nothing else to protect against CVR upload issues, bugs in the code etc. That is a standard part of an RLA, but it is in some sense separable from the data gathering that the tool is doing.
We could do the calculations as a check during the importing of the data, without storing the CVRs in that sort of database table. I see that as easier than parsing the Dominion tabulation results file. We already have `rla_tabulation_outcome` for storing the tabulation outcomes, which should also have margin information.

dmzimmerman: Makes sense; we just need to figure out the best way to handle it. We also need to figure out how we're getting election definitions into the system, but that's a whole other thing.

nealmcb: In terms of UI, we could provide our tabulation information as a result of them uploading the CVRs, and share that publicly and with the county and sos for them to check against the announced results.  At some point, the public has to know the tabulation the tool is relying on anyway.
These checks are all part of the AXIOM TBDs in the "CVR Export Verification" section that I commented on long ago. At least I think I did.....

nealmcb: The election definitions that we need (I think) are all in the CVR column header lines, e.g. `Presidential Electors (Vote For=1)`.

dmzimmerman: Ah.  OK, so the import process is basically going to be importing both the election definitions and the CVR data at the same time.

nealmcb: Yes :) That's the way I've had to handle this stuff since 2008 in `electionaudits`.

nealmcb: In the future, we could work with the `county_contest` info that we actually have to make up a `contest` table which includes all `county_contests` that a particular contest involves, matching on the contest `name`.
We've been working on them for months, as have they, to ensure that the naming of contests is uniform across counties. We'll see if that actually works....

dmzimmerman: I believe it is "convert a single CVR to a line for inclusion in a CSV file".

dmzimmerman: This seems reasonable to me.

specs/pvs/corla.pvs
dmzimmerman: It looks like it is, but I'm not sure I see the utility of it; it seems to me that some identifier for the election in which the contest appears could be passed to `select_contest_for_comparison_audit` instead of maintaining this state in a session here.

specs/pvs/corla.pvs
nealmcb: Thanks for digging in to an important topic. It is important for each contest to ensure that the samples we're considering were acquired uniformly and at random. To start out with, we can take the easy approach of only paying attention to a single county and a single contest that is only in it, sample all ballots in the county, ensure that we're looking at all ballots selected in order for that county and ensure we have an answer for each one up to some high water mark, and thus know that our sample is uniformly random.  Some day we will support more efficient sampling of sub-county contests, but we're not there yet.
One really tricky part is that we can't trust the CVRs for knowing exactly which ballots are on which contests. They are what we're auditing and they might not be right. Harvie is working hard on getting independent evidence in the manifest for the ballot style of each ballot, but that's a tall order.

specs/pvs/corla.pvs
nealmcb: Ahh right.  I was thinking that this was related to establishing interest in auditing that election in the first place, but I guess we prefer that the client be as stateless as possible.
But doesn't that mean we'll need endpoints to request lists of contests and maybe elections to choose from?
And the "establish risk limit" will I think need to apply to each individual contest, unless we have some other means for reading them in. Could/should setting it be part of the selection of the contest for audit instead of being a separate step?
And "_publishing the list of ballots to be audited for each audited contest_" seems like it should be per county as I just discussed elsewhere.

specs/pvs/corla.pvs
kiniry: I was planning for the later addition of other kinds of risk-limiting audits, since the C.R.S. says that they can/ determine both audit type and risk-limits based upon a variety of factors.

specs/pvs/corla.pvs
kiniry: It is supposed to be "selecting contest" in the comment, of course.  As contests relate to elections, there is no need to pass an election.  The refresh endpoint provides the lists you mention, @nealmcb.  

At the state level CDOS has asked for a list of ballots to be audited in each contest (possibly spanning counties).  At the county dashboard level, they'll see the (filtered) list of ballots for their county.

kiniry: I meant in the audit board dashboard, of course, not the county dashboard.

specs/pvs/corla.pvs
kiniry: Ok.

specs/pvs/corla.pvs
kiniry: Hrm, I have been usually going with the singular on type names even for types that are defined as sets of Ts.  But I'm not particularly bothered by this if it helps with clarity and comprehension.  I'll make the change.

kiniry: I'm not entirely clear what it is for either.  It is a term used in the draft C.R.S. which I think means the *uncorrected* tabulation outcome that is used as an input to the RLA algorithm.

kiniry: Yes, given recent information from CDOS, this should be refined to a list (which is a finite sequence).

kiniry: Indeed, now that we have sample CVRs from CDOS, this definition could be enormously enriched, though we usually stick to modeling only exactly those subterms that are necessary to define the system's behavior.  I'll make the change that you note, but we should realize that it'll likely evolve more soon.

dmzimmerman: Yes, it should be a sequence I would think.

dmzimmerman: Note that `ballot_manifest_info` currently does not have `ballot_id` in it, but it should, right?

kiniry: @dmzimmerman is correct.

kiniry: @sfsinger19103 and I were talking about annotating the domain model to highlight which concepts are necessary for the RLA Tool that we are building for this contract, and which ones are there for other reasons.  E.g., some terms are there because they are part of the domain model of elections in the USA.  Others are there because they are mentioned in the C.R.S. but are not (at the moment) mentioned in the Section on RLAs.  What do you all think about such?  I intend to write a paragraph about this in the preface anyway (there is a `@todo` for such), but highlighting them in an extra special fashion (like special coloring, margin comments or markers, etc.) is another more detailed way to go. @mkiniry can help with this polishing, even as a part of his current work in the `glossary-specification-double-check` branch.

nealmcb: The model we've mostly discussed for the general case in CORLA2 would still have just one table ala `county_contest_table`, though perhaps it could be called `collection_contest_table` to be more general, following Ron Rivest's terminology for the smallest jurisdiction that maintains a actual physical collection of ballots and can inspect a given sample of them. Contests can involve a subset of that collection, or involve several collections, all the way to the United Federation of Planets spread across all the galaxies, to the extent that we could communicate fast enough...

dmzimmerman: @kiniry I think this would be an excellent idea because it would help us narrow down what exactly needs to be implemented as Java classes; in particular, @jldodds and I are experiencing some confusion at the moment as we try to implement `CastVoteRecord` and `Vote`.

dmzimmerman: Is a CVR going to _contain_ a `ballot_manifest_info`, or be associated with one in a table row? (recall that a `ballot_manifest_info` is itself a set of something like 6 different data items, so while we could index on that, it might be cumbersome.

nealmcb: We should limit what we index on to something that contains just the essential key they're using: "TabulatorNum", "BatchId", "RecordId". We should just drop stuff that shouldn't be in there anyway for CORA reasons: _These columns seem problematic: "ImprintedId", "CountingGroup", "PrecinctPortion".  And in this case the "BallotType" (ballot style) seems to include a precinct number, which is also a problem given the way we mix ballots up in batches in Colorado._ The CvrNumber is redundant, just seems to be another unique incrementing recordid.

nealmcb: Ah yes, and in terms of whether it is contained or associated, I don't have an opinion offhand.

sfsinger19103: FWIW, CDOS folks seemed to say on the call today that the Dominion BallotType field corresponded one-to-one to the subsets (of the set of all contests) that actually appeared on ballots in the election. The example they gave was that if there was only one contest in a particular election, and if it was a county-wide contest, then there would be only one BallotType. If this is not true, a counter-example would be helpful.

dmzimmerman: So then maybe a CVR should not contain a `ballot_manifest_info`, but should contain those other essential fields (the ones it ships with in the CSV file). "ImprintedId", I thought, was just "TabulatorNum-BatchId-RecordId" - that's what it appears as in the CSV files, and I'd imagine it would be fine to just index on that. Why do you think it's problematic?

dmzimmerman: And if it doesn't contain a "BallotType", how do we figure out what contests to list on the auditor dashboard for that ballot when it gets randomly selected?

nealmcb: Hmm - looking again (at my cleaned up version of the example CVR that github can show in table form: https://github.com/nealmcb/ColoradoRLA/blob/cleancvr/test/dominion-2017-CVR_Export_20170310104116-clean.csv) I don't see what I thought I saw last week. Might have confused it with the previous field (PrecinctPortion). But the file is so abbreviated it's hard to know. Anyway if they say its a 1-1 mapping, that's what we want. Thanks, and sorry for the confusion. Indexing on ImprintedId would seem to get things out of order if it is a string and we want stuff in order by numeric tabulator num etc.

dmzimmerman: Well, indexing (in the database sense of index) and sorting don't necessarily have to be done on the same columns. But yes, that makes sense.

kiniry: I admit to being confused.  It was my understanding that sampling always happens across an entire contest, independent of geography.  Thus, the type of this model concept only mentions the product of contests and sequences of ballot contest info.  I intended to filter off of this sequence based upon county.

kiniry: `cvr` now contains `ballot_manifest_info`.

When you say "index .. by", what do you mean in the modeling context?  Do you mean that an order can be defined on the non-indexed elements, or only that there is an injection from the index to the remaining projections?

dmzimmerman: That was my understanding too, Joe.

kiniry: Note that this is a data model, which is only meant to characterize relationships between types, not how they are organized or stored.  Perhaps this conceptual perspective is the source of the confusion I have about some of your earlier comments.   The only point of this concept is to (conceptually, not necessarily actually) aggregate all CVRs uploaded by counties.

nealmcb: Thanks for clarifying your intention - I missed that. The first problem is that most folks want sampling to happen from the manifests, which don't have ballot-by-ballot style info. We might be able to finesse that as noted in a separate conversation. But still, if we do sampling by contest, wouldn't we have disjoint samples in the same county for different contests? We want a single sample to cover all audited contests. CORLA2 has always framed it in terms of sampling by county, and maybe someday doing so by CVR to optimize finding contests that don't appear on all ballots, but that last part is tricky since we can't trust the CVRs.

kiniry: This is the relation between each contest and its audit report.

kiniry: I removed `ballot_manifest_info` in my recent commit (9994328).  In this spec, I am, in fact, using ballot to represent the encoding of the auditors, so it is wise for us to add a new type (aCVR is fine with me) to represent that information.  Go ahead and add timestamps to the model too.  Right now logging is completely underspecified because we do not really know what we must log (my contract and law) and what we should log (for forensic auditing and debugging).

kiniry: I certainly do not really care how the DB data model gets refined from what is in this spec, so long as each persistent datatype is realized (without fidelity loss) and each relation (as encoded in all of the `*_table` types) is obtainable via query.  Some of that encoding is going to be undoubtedly driven by whatever awkward schema you end up reverse engineering from Dominion resources.

kiniry: Yes, I think it should.  Note the commented out specification of ballot because of our lack of information about Dominion schemas when I wrote that module.

kiniry:
I'm not touching this spec anymore and believe I have addressed all reviews.  @ranweiler and @nealmcb should check that, sign off on the branch, and you all can do the merge.

kiniry:
I have rebased and merged to make the merge with master easy.

dmzimmerman:
@kiniry, I don't think you really rebased, as there are a number of commits on master that didn't make it here... I can do that if you want, but I'm not going to do it tonight because I'm heading to bed soon.

specs/pvs/corla.pvs
nealmcb: `summary_results_table` summarizes a Dominion report about overvotes, blanks, and valid votes. I guess in some sense that is Dominion's version of an audit report, though the term is confusing in our context. It certainly would be a useful thing for us to audit that report against the CVRs etc.
But I haven't seen any requirements around us dealing with their report. So I guess it is an opportunity for future work.

nealmcb: At some point it would seem useful to  distinguish between a `csv` (the whole file: headers defining columns, rows) and a `csv_row` (a single row). Or perhaps that isn't needed, if we spec this as a conversion from a sequence of cvrs to a csv.
`  as_csv: [ sequence[cvr] -> csv ]`

nealmcb: For clarity, I'd drop these. Otherwise questions arise about the need for literally hundreds of other tables like st_vrain_school_district_contest_table, which spans county boundaries but doesn't seem appropriate for state or national.
Accommodating the highly complex nature of Colorado district boundaries is a key concern for election software here.

nealmcb: In the future, the RLA calculations will presumably be based on separate samples for each county, which will probably happen with a different sampling rates in each county.
For now, `ballots_to_audit_list_table` sounds like the list that is presented to the county before they actually audit them, thus my confusion.  Changing the name to something like `audited_ballot_list_table` might make it clearer. But I'm good with this for this purpose and phase.

nealmcb: Since we won't get a list of ballot marks, I'd drop that, and at some point change `ballot` to an `acvr` defined just like a cvr, (or perhaps that can just be a second `cvr` here?). Finally add one `modification_timestamp` (do you just say `string` here? or `modification_timestamp: string`?), defined as the last modification time.

Note that audit board input like "ballot doesn't exist" need to be encoded as an acvr with votes for all the losers (a `phantom_ballot`). I'm not sure how/where that is specified or implemented (client or server?). Ahh, perhaps that is why kiniry was specifying `ballot` here in the first place, along with the list of marks.

nealmcb:
@ranweiler are you ready to approve your requested changes?
I updated my summary of requested changes, [just above](https://github.com/FreeAndFair/ColoradoRLA/pull/108#pullrequestreview-52071440). Still two that I think should be updated.

specs/pvs/corla.pvs
```diff
@@ -2022,7 +2116,209 @@ BEGIN
...
+%%% Each feature is encoded in a function in the
+%%% \texttt{department\_of\_state\_dashboard} module.
+
+%%% \begin{enumerate}
+
+%%% \item \texttt{authenticate\_state\_administrator}: authentication
+%%% of state administrators (the mandatory first step in the
+%%% protocol),
+
+%%% \item \texttt{select\_contest\_for\_comparison\_audit}: defining
```
ranweiler: This item name, "select contest for comparison audit", is still accidentally repeated below.

specs/pvs/corla.pvs
nealmcb: Addressed in 1dd1860

kiniry: You can just say `string` for now.  We should start to use the `Time` type from the PVS prelude.

specs/pvs/corla.pvs
kiniry: Got it.  I have removed the duplicate item and fixed the name on its sister.  

I have also removed all unicode punctuation.

Please review @ranweiler so we can get this merged.

specs/pvs/corla.pvs
kiniry: To remedy this discussion and provide us a means by which to reason about ballots under audit from multiple perspectives, I have renamed the original table mostly per @nealmcb's recommendation (relating counties to sequences of ballot manifest info) and I have added another table to continue to capture the old relationship to contests.

kiniry: Yes, I just wanted to attempt to capture what a standard summary report looks like. If/when we can, we'll refine it to accurately capture that which different vendors' equipment emits.

kiniry:
@nealmcb I believe I have addressed all of your suggestions in 24ab032.

kiniry:
I had to do a little merge on a2d2762.

specs/pvs/corla.pvs
```diff
@@ -2056,9 +2055,15 @@ BEGIN
...
     sequence[[RLA, audit_reason, contest, risk_limit]]
-% @todo nealmcb rename to audited_ballot_list_table?
-  ballots_to_audit_list_table: TYPE =
+%%% We formalize the list of ballots under audit and organize them
+%%% both by contest and by county. The former is defined when the RLA
+%%% algorithm randomly chooses ballots. The latter is populated by
+%%% filtering the former by county, and it is what is shared with
+%%% counties via the Audit Board Dashboard.
+  audited_ballot_by_contest_list_table: TYPE =
     sequence[[contest, sequence[ballot_manifest_info]]]
```
nealmcb: Per discussion with @dmz, we don't choose by contest. We choose by county and stratify by county across contests.
We will probably end up filtering by contest from that. But that is very tricky to do since we don't trust the CVRs. So we need independent-of-the-voting-system ballot manifest evidence of counts by contest, which we don't have yet.
This affects multi-sheet ballots since they are provided as one CVR per ballot sheet it seems.

nealmcb: And we also have to choose by county, not contest, so we end up selecting the same ballots for each contest.
Proposed text:
```
%%% We formalize the list of ballots under audit and organize them
%%% both by county and by contest. The former is defined when the RLA
%%% algorithm randomly chooses ballots, and it is what is shared with
%%% counties via the Audit Board Dashboard. The latter is populated by
%%% filtering the former by contest, for future use.
```
and then swap the order of the definitions.

kiniry:
That just took ~30 manual merges.  Let's not do *that* again.

PR #114, opened 2017-07-25
Fix two latex errors for pvs make

nealmcb:


nealmcb:
Note: this is a merge request into `protocol-spec`, the branch for the corla.pvs branch #108

PR #119, opened 2017-07-26
Server development for progress toward initial deliverable

dmzimmerman:
Implementation of the server side endpoints, model classes and database connections.


dmzimmerman:
I believe this is finally ready to be merged into master. I will of course rebase it before doing so.

dmzimmerman:
Hmm. Static analysis on the command line is still unhappy. Investigating now...

PR #139, opened 2017-07-28
Client Phase 1 sprint

ranweiler:


client/src/example.ts
```diff
@@ -87,9 +102,18 @@ export const exampleCountyState = {
     ballots,
```
nealmcb: The date examples **could** be tweaked, if the client will see this:
```
    electionDate: '11/21/2017',
        auditDate: '12/1/2017',
```
to
```
    electionDate: '11/07/2017',
        auditDate: '11/18/2017',
```

NO LATER THAN 5:00 P.M. MT ON THE
32 FRIDAY AFTER ELECTION DAY, THE SECRETARY OF STATE WILL SELECT FOR
33 AUDIT AT LEAST ONE STATEWIDE CONTEST, AND FOR EACH COUNTY AT LEAST
34 ONE COUNTYWIDE CONTEST.

ranweiler: I will change this as you suggest @nealmcb, thanks!

ranweiler: Done in 6d94f96.

PR #146, opened 2017-07-28
Endpoint for upload of audit CVRs.

dmzimmerman:


ranweiler:
The only issue I see here is that we want to track both a `comments` field for each contest under audit, and and an ability to note that the audit board could not come to a consensus. Would you like to do that here, or a separate PR?

dmzimmerman:
Hmm. A separate PR would work for that. It's easy enough to add another table (contests to comments) to the CastVoteRecord class.

PR #166, opened 2017-07-31
Persistence and Large File Uploads

dmzimmerman:
This includes eliminating final fields, adding annotations for appropriate data persistence semantics, and working out how to manage database/entity management sessions.


dmzimmerman:
Persistence to PostgreSQL works now, at least in conjunction with the static HTML test page (which is updated now to reflect that BallotStyle is not a downloadable thing of its own). 

If you want to test persistence, you must:

1. Install PostgreSQL (brew install postgres, apt-get install postgres, or whatever is appropriate) and start it running.
2. Create a database called "corla", and grant all privileges on it to a user called "corla" with password "corla"
3. That's it. If the database is there the server will use it (and will, at this stage, create all its tables and such automatically); if not, it will use in-memory persistence.

dmzimmerman:
This is, I think, ready to be merged.

dmzimmerman:
(Pending review, of course)

nealmcb:
Re: postgreSQL names and versions. `apt-get install postgres` isn't a thing as far as I see.

It seems that RHEL/CentOS/SL/OL 7 supports PostgreSQL version 9.2 (also supplies package rh-postgresql95 and rh-postgresql94 via SCL)

It looks like the "supported" version on trusty that comes via `apt-get install postgresql` is 9.3+154ubuntu1 which seems pretty close.  I'm no mac user, but "postgresql" might work with brew as a more canonical name than just "postgres".  It also might be a more recent version like 9.6.

Are there any version dependencies we're aware of, or versioning we want to specify?

kiniry:
I have added some docs for the postgres setup to `25_developer.md` in the `wire-format-review` branch I'm working on.

kiniry:
Homebrew installs 9.6, FWIW.

server/eclipse-project/src/main/resources/us/freeandfair/corla/default.properties
```diff
@@ -10,3 +10,8 @@ hibernate.url = jdbc:postgresql://localhost:5432/corla
 hibernate.user = corla
 hibernate.pass = corla
 hibernate.dialect = org.hibernate.dialect.PostgreSQL9Dialect
+hibernate.c3p0.min_size = 5
+hibernate.c3p0.max_size = 20
```
dmzimmerman: I haven’t even checked to see that c3p0 works for connection pooling automatically yet without further configuration, but I had seen 20 in the “how to set this up” docs I read about it. In reality I’d be shocked if we even need a connection pool half that size. But if the default is 100 now it might be worth sticking to that.

dmzimmerman:
OK, rebased against master; let's merge it.

PR #170, opened 2017-08-05
Large file uploads

dmzimmerman:
Enabling large file uploads in the database.


dmzimmerman:
To fix: return values from File.delete, stream resource in UploadedFile.

To do: change the parsers and endpoints that download JSON to be streaming and not hold tables of persistent resources.

dmzimmerman:
Folded back into #166

PR #172, opened 2017-08-06
Archive Deliverables in git?

nealmcb:
Should we archive the deliverable zip files in github, for convenient access by testers etc?
Or somewhere else?
For now, I've put the first two in the test directory, but I'm happy to rearrange as requested.


ranweiler:
I strongly recommend against this. This works against what Git (not -Hub) is good at and intended for.

If we are going to use Git*Hub* for distribution, the nicest way would be via a "release": https://help.github.com/articles/creating-releases/

Note steps 7 (attaching binaries, we can use this to distribute zip files) and 8 (we can and should flag the early deliverables as pre-release).

Separately (and less emphatically), I'd suggest making the next deliverable our first GH "release", since it will have an integrated client and server. I think the existing distribution is fine for the Phase 1 deliverable. Also, since this is also an open-source project, I think we should strive to separate the idea of "deliverable" (a contractual notion) from "release" (an engineering notion), and name files after the system being released, not their contractual role.

kiniry:
I agree with @ranweiler's input on this matter moving forward.  But archiving these two first deliverables should happen somewhere for posterity, if nothing else.

dmzimmerman:
I agree as well. And I believe they are in Google Drive at the moment

nealmcb:
Thanks. I knew I'd get good feedback here!
It sounds like the consensus is to preserve our own copies of these deliverables, rather than releasing them, and to do a release in parallel with the next deliverable. Makes good sense.
For now I'll close the PR, but leave this branch on GitHub, since I already responded to some folks who were interested in trying it out, and pointed them this way, while noting it was a temporary location.  I'll delete the branch when the release comes out since that will indeed be much better.

PR #175, opened 2017-08-07
Tutorial edits

sfsinger19103:
A first version of the tutorial.


docs/20_tutorial.md
```diff
@@ -14,10 +15,161 @@ the computers read those ballots, to determine whether the computers
...
+generate and preserve
+a Summary Results Report and a Results File Export, the RLA Tool 
+does not interact with these files in any way. Nor does the RLA Tool 
+aid in any way with the CVR Export Verification required by the 
+RLA process. 
+
+While the RLA Tool does provide for upload of the 
+hashed Ballot Manifest, the hashed CVR file and their hashes, 
+the RLA Tool does not provide a hashing utility.
+***review please!
```
kiniry: This is correct.  We *check* they hash that they provide.

docs/20_tutorial.md
```diff
@@ -14,10 +15,161 @@ the computers read those ballots, to determine whether the computers
...
+
+The RLA Tool does not provide any support for paper ballot retrieval or
```
kiniry: This isn't entirely true.  On the audit board dashboard we can/will display the ballot manifest information for the selected ballots under audit, including scanner id, batch id, and the storage container that should contain the ballot in question.  I'm right about this, @dmzimmerman @MorganMillerUX @nealmcb, aren't I?

docs/20_tutorial.md
```diff
@@ -14,10 +15,161 @@ the computers read those ballots, to determine whether the computers
...
+## For County Audit Boards
+========
+
+### Signing In
+
+The Audit Board does not log directly into the RLA Tool with
+computer system usernames and passwords. However, whenever 
+the Audit Board begins to interact with the RLA Tool, either 
+at the beginning of an audit round or after taking a break, 
+the Audit Board must sign in.
```
kiniry: Explain what "sign in" means.  The County Administrator has to *authenticate*.  The County Audit Board members have to informally "sign in".

docs/20_tutorial.md
```diff
@@ -14,10 +15,161 @@ the computers read those ballots, to determine whether the computers
...
+
+Click on "Start Audit"
+
+Screenshot (page:Audit Board Sign In) ---
+
+Click "Next"
+
+### Auditing Ballots
+
+Note: do *not* use your browser's "back" arrow during the audit process.
```
kiniry: Is this really true, @ranweiler?  It'd be unfortunate if accidental use of the back button screwed audit state.

docs/20_tutorial.md
kiniry: Oh, and @ranweiler ...

docs/20_tutorial.md
```diff
@@ -14,10 +15,161 @@ the computers read those ballots, to determine whether the computers
...
+
+Screenshot (Ballot Verification) ---
+
+Before recording voter intent, please double-check that the paper ballot
+ID and Ballot Style match the ID and ballot style listed on the screen.
+Then record on the screen all final voter intent marked on the paper ballot.
+In case of an overvote, mark each of the (too many) choices the voter
```
kiniry: This descriptive text will need to be carefully reviewed by CDOS folks to make sure it aligns in terminology and intent with their Rules.

docs/20_tutorial.md
```diff
@@ -14,10 +15,161 @@ the computers read those ballots, to determine whether the computers
...
+Note: do *not* use your browser's "back" arrow during the audit process.
+
+Screenshot (Ballot Verification) ---
+
+Before recording voter intent, please double-check that the paper ballot
+ID and Ballot Style match the ID and ballot style listed on the screen.
+Then record on the screen all final voter intent marked on the paper ballot.
+In case of an overvote, mark each of the (too many) choices the voter 
+intended; the RLA tool will recognize the overvote. In case the Audit 
+Board cannot reach consensus, mark the "No Consensus" box on the screen.
```
kiniry: ...and write a short explanation of the reason for the audit board's lack of consensus in the `Comments` text field.

docs/20_tutorial.md
```diff
@@ -14,10 +15,161 @@ the computers read those ballots, to determine whether the computers
...
+
+Screenshot (Review Screen) ---
+
+Check that the information on the review screen reflects the 
+Audit Board's interpretation of the ballot. Note that in case of 
+an overvote, the review screen will say, simply, "overvote".
+
+Note: do *not* use your browser's "back" arrow during the audit process.
+If the review screen does not match the Audit Board's interpretation,
+click "Back".
```
kiniry: Sorry, but I think that this is unacceptable UI/UX, @ranweiler and @MorganMillerUX.

docs/20_tutorial.md
dmzimmerman: Yes, I believe you are right about this. Certainly we have that information and can retrieve it.

docs/20_tutorial.md
sfsinger19103: I view the Comment text field as purely optional, a way to collect anything that the Audit Board feels is not addressable through the rest of the interface. 

What value would be added by encouraging an explanation of reason for lack of consensus?

docs/20_tutorial.md
sfsinger19103: I agree.

docs/20_tutorial.md
ranweiler: Just saw this. I believe this is an overstatement: the worst the back button would do right now is take them to the wrong page, if they hit it during the audit wizard. If they navigate back to the correct URL path, they will see the correct thing. The browser back button works generally. We will change the wizard to be fully browser-back friendly, though.

docs/20_tutorial.md
kiniry: Primarily traceability of the audit, right @nealmcb?  I can imagine that such comments might even end up being used as evidence in a court of law someday, e.g.

docs/20_tutorial.md
ranweiler: Agreed. Per my comment above, we can change this.

The reason it is like this right now is because the "wizard" is implemented as a component that changes at one particular URL path, `/county/audit`, with the current wizard view/"stage" determined by an internal variable. We can change this to make it so the active wizard stage match a further path part, e.g. `/county/audit/review`, or similar.

docs/20_tutorial.md
sfsinger19103: Hm, the price is slowing down the Audit Board. They'd have to come to consensus on what to write in the box, seems like a can of worms to me. Curious to know what @nealmcb thinks.

nealmcb: I think they should be encouraged to write an explanation in the comments box. If they refuse, so be it. If is slows them down to much just discussing what to write, they will decide if its worth it or not to continue. I don't think we should pretent to tell them what to do, just explain how they can use the tool. So I'd just update kiniry's proposed text to something like "You can write a short explanation of the reason for the lack of consensus in the Comments text field."

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+The Risk-Limiting Audit (RLA) tool, developed by Free & Fair 
+for the Colorado Department of
+State for use in elections from November 2017 forward, supports a 
+Risk-Limiting Audit as required by Colorado statute and 
+as described in the Colorado Secretary of State's Rule 25 for
+elections. RLA-related terms and processes are defined and prescribed in Rule 25.
+*** citation for Rule 25.
+
+The RLA Tool enables the Secretary of State and each individual 
+County to carry out a Risk-Limiting Audit. The Secretary of State can
```
nealmcb: I think we've been putting that in lower case: risk-limiting audit.

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+enter the risk limit, the random seed and the contests that drive
```
nealmcb: > risk limits

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+the audit process. The Secretary of State can also identify 
+contests that should go to a full hand count. Counties can upload
```
nealmcb: I'd defer mention of full hand counts until somewhere buried deep in here after describing rounds etc.

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+ballot manifests and cast vote record (CVR) files. The
+RLA tool performs the necessary statistics (such as
```
nealmcb: "necessary statistics" => "necessary calculations"

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+an overvote, the review screen will say, simply, "overvote".
+
+Note: do *not* use your browser's "back" arrow during the audit process.
+If the review screen does not match the Audit Board's interpretation,
+click "Back".
+
+If the review screen matches the Audit Board's interpretation,
+click "Submit and Next Ballot". This submission is final. There 
+is no way to revise a ballot interpretation once it has been submitted
+from the Review Screen.
```
nealmcb: => "there is currently no way" or "in the current version there is no way"

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
 Besides a printed document with screenshots and instructions, we could also
 offer to create a tutorial video (videos, probably) where we film people using
 the tool/dashboards, with narration describing what they are doing, and what comes
 next depending on what’s happening.
+***SS says, it would be nice, but out of scope. We're so far over budget already....
+
+Brief introduction to RLAs in non-technical, human friendly language.
+At the county level, something along the lines of "This tool is going
+to help you cross reference votes recorded on paper ballots with how 
+the computers read those ballots, to determine whether the computers
```
nealmcb: "how the system interpreted those ballots". Remember, interpretation is not purely done by computers. Human adjudication is part of the system.

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+did their job accurately…” Try to put their minds at ease about the
```
nealmcb: "how the system interpreted those ballots", and "whether the system performed accurately". Remember, interpretation is not purely done by computers. Human adjudication is part of the system.

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+aid in any way with the CVR Export Verification required by the 
+RLA process. 
+
+While the RLA Tool does provide for upload of the 
+hashed Ballot Manifest, the hashed CVR file and their hashes, 
+the RLA Tool does not provide a hashing utility.
+
+The RLA Tool does not provide any support
+for chain of custody logs. Nor does it provide any guidance for ballot 
+interpretation.
```
nealmcb: "does not currently"...

docs/20_tutorial.md
sfsinger19103: For the User Manual I agree. OK with you if the Run Book is more brass-tacks, aimed not to teach people about RLAs but to enable people involved in RLAs (and so presumably at least a bit knowledgeable) to use the computer tool to do what they need to do?

nealmcb: That sounds like the right contrast. But this is the intro of the document that people will be more likely to read, and I just don't want to scare them off at the beginning with talk of full hand counts. So I'm just suggesting that we could say here in the intro something like "and all other necessary functions", and leave the full hand count details for the appropriate places.

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+Although the RLA process requires Counties to 
+generate and preserve
+a Summary Results Report and a Results File Export, the RLA Tool 
+does not interact with these files in any way. Nor does the RLA Tool 
+aid in any way with the CVR Export Verification required by the 
+RLA process. 
+
+While the RLA Tool does provide for upload of the 
+hashed Ballot Manifest, the hashed CVR file and their hashes, 
+the RLA Tool does not provide a hashing utility.
```
sfsinger19103: Created issue #197

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+
+Screenshot (SoS authentication) ---
+
+### Identifying County Users
+
+The Secretary of State will determine, for each individual
+County, which user will be able to log in to the County-facing
+part of the RLA Tool.
+
+Screenshot (SoS County User Identification)
```
sfsinger19103: When the front end is done we'll review to make sure this kind of thing matches.

docs/20_tutorial.md
```diff
@@ -1,64 +1,245 @@
...
+Once the ballot manifest and CVR files are verified and hashed, 
+they can be uploaded.
 
-Detailed description with associated screenshot of the County dashboard
-with instructions about the process of uploading ballot manifests and
-CVRs, and associated hash functions.
+Screenshot (page:county) ---
+
+Note that the RLA Tool does not support export of the
+tabulation results to the Election Night Reporting System.
```
sfsinger19103: Paragraph 1: you and I have different concepts of the audience and purpose of the Run Book. It's a limited, streamlined document, simply designed to guide the people who have to interact directly with the computer system. It's not going to be read cover to cover but used as a reference as people learn to use the computer system.

Paragraph 2: point taken.

Paragraph 3: Out of scope for the Run Book. See Paragraph 1.

PR #184, opened 2017-08-07
Massive cleanup of the persistence and model classes

dmzimmerman:
This includes:

- elimination of local caching
- model classes are now actually model classes and don't know anything about the database or Hibernate other than that they are JPA entities (by virtue of annotations)
- adding new entities is significantly more straightforward
- basic queries are available from the Persistence class, more complex queries are done only by endpoints as needed
- many endpoints actually use HQL to do their queries now
- database transaction management now extremely easy to work with

Still to do (but perhaps not this week):

- stream results from the database, for large queries, to avoid running out of memory once we're dealing with gigabytes of CVR data
- improve the efficiency of some queries that are being done, right now, using query by example


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRUpload.java
```diff
@@ -84,14 +87,16 @@ public String endpoint(final Request the_request, final Response the_response) {
...
                                   acvr.recordID(), acvr.imprintedID(), acvr.ballotType(), 
-                                  acvr.contestInfo());
+                                  acvr.contestInfo()),
+                          CastVoteRecord.class);
       Main.LOGGER.info("Audit CVR parsed and stored as id " + real_acvr.id());
-      Main.LOGGER.info(CastVoteRecord.getMatching(new HashSet<String>(), 
-                                                  RecordType.AUDITOR_ENTERED).size() + 
-                       " audit CVRs in storage");
+      final long count = count();
+      if (count >= 0) {
```
ranweiler: I would prefer we used [`Optional<Long>`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) for methods like this.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestUpload.java
```diff
@@ -93,7 +95,7 @@ private UploadedFile attemptFilePersistence(final File the_file,
                                               final String the_hash) {
     UploadedFile result = null;
     
-    if (Persistence.isEnabled()) {
+    if (Persistence.hasDB()) {
```
ranweiler: This appears in several methods. It would be nice if we could factor it out to a higher level. Failing that, even using an early return would be nice, in that it would reduce the level of conditional nesting.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestUpload.java
```diff
@@ -198,8 +200,10 @@ private void parseAndPersistFile(final UploadInformation the_info) {
         }
         Main.LOGGER.info(parser.parsedIDs().size() + 
                          " ballot manifest records parsed from upload file");
-        Main.LOGGER.info(BallotManifestInfo.getMatching(null).size() + 
-                         " uploaded ballot manifest records in storage");
+        final long count = count();
+        if (count >= 0) {
```
ranweiler: The above comment about `Optional` applies here. It seems like this is shared in several places, too. Maybe we should factor it out, or at least make them call a parameterized common method?

server/eclipse-project/src/main/java/us/freeandfair/corla/hibernate/Persistence.java
```diff
@@ -181,148 +195,250 @@ private static synchronized void setupSessionFactory() {
...
+   * @return true if a long-lived transaction is running in this thread, 
+   * false otherwise.
+   * @exception PersistenceException if the database isn't running.
+   */
+  public static boolean isTransactionRunning() 
+      throws PersistenceException {
+    if (hasDB()) {
+      final Pair<Session, Transaction> session_transaction = transaction_info.get();
+
+      return session_transaction != null &&
```
ranweiler: This logic is repeated (essentially) on a bunch of lines. Can we factor it out into a method or function whose name would also summarize the conditions being checked?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -60,18 +65,58 @@ public String endpointName() {
...
+  private Set<CastVoteRecord> getMatching(final Set<String> the_county_ids) {
+    Set<CastVoteRecord> result = null;
+
+    try {
+      final Set<CastVoteRecord> query_result = new HashSet<>();
+      Persistence.beginTransaction();
+      final Session s = Persistence.currentSession();
+      for (final String county_id : the_county_ids) {
+        final Query<CastVoteRecord> query = 
+            s.createQuery("from CastVoteRecord where county_id = '" +
```
dmzimmerman: Queries aren't likely to stay created this way anyway. I was intending to convert them to type safe JPA queries rather than writing any SQL by hand, I just hadn't gotten to that yet.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownload.java
```diff
@@ -56,18 +62,50 @@ public String endpointName() {
...
-      bw.flush();
-      return "";
+      final Set<CastVoteRecord> matches = getMatching();
+      if (matches == null) {
+        result = "Unable to fetch records from database";
+        status = HttpStatus.INTERNAL_SERVER_ERROR_500;
+      } else {
+        Main.GSON.toJson(matches, bw);
+        bw.flush();
+        result =  "";
```
dmzimmerman: The coding guideline at play is "single exit point where possible".

ranweiler: :+1: will keep in mind in future feedback.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -50,7 +55,7 @@ public EndpointType endpointType() {
    */
   @Override
   public String endpointName() {
-    return "/acvr/:counties";
+    return "/acvr/county/:counties";
```
dmzimmerman: I'm happy to change that. Seems pretty straightforward to do... but what would `POST /acvr/county/11` mean then? Actually uploading an ACVR for that particular county rather than submitting the county's name as a parameter to the upload?

server/eclipse-project/src/main/java/us/freeandfair/corla/hibernate/Persistence.java
dmzimmerman: I'll take a look at that.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestUpload.java
dmzimmerman: I have not yet really adopted `Optional` into my Java coding; maybe I should, though.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRUpload.java
dmzimmerman: Perhaps `OptionalLong`, since it exists and we don't need the object wrapper.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
ranweiler: Ah, `PUT` or `DELETE` is a better in this case, but in others it often makes sense to `POST` data to a particular entity to create a subresource. I probably just shouldn't have chosen the exact worst HTTP method to illustrate my point (-:

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
ranweiler: Ah, that makes sense. Well I'll let you judge whether to do a quick fix now or wait for the more meaningful one, though I'd prefer to do SQLi-friendly interpolation at all, even temporarily. Time is precious, of course, so I'll defer to you.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRUpload.java
ranweiler: Ah, even better.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestUpload.java
ranweiler: Yeah, I mean, that's fair, because if we were on 7 it wouldn't even be a choice. Might be nice for us, though.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestUpload.java
dmzimmerman: This actually made me notice that I had failed to clean up `UploadedFile`, so thanks for that. Though I'm not sure my fix will make you thrilled. :)

server/eclipse-project/src/main/java/us/freeandfair/corla/hibernate/Persistence.java
dmzimmerman: I don't think that they are similar enough to make it worth factoring out; I'd need different methods for the different sets of conditions, or a way to specify to the single method which set to check, which boils down to the same thing.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
dmzimmerman: OK. Well, I'll change the endpoint for the get, anyway (in all the places where lists of county names appear - by the way, they are county names, not county IDs, in the current implementations).

server/eclipse-project/src/main/java/us/freeandfair/corla/hibernate/Persistence.java
ranweiler: Fair. I guess there's probably no easy way to be polymorphic in `&&`/`||` and get the same short-circuiting behavior, which definitely makes it not worth it.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRUpload.java
```diff
@@ -93,8 +94,8 @@ public String endpoint(final Request the_request, final Response the_response) {
                                   acvr.contestInfo()),
                           CastVoteRecord.class);
       Main.LOGGER.info("Audit CVR parsed and stored as id " + real_acvr.id());
-      final long count = count();
-      if (count >= 0) {
+      final OptionalLong count = count();
+      if (count.isPresent()) {
         Main.LOGGER.info(count + " ACVRs in storage");
```
ranweiler: This needs to become `count.getAsLong()` now, doesn't it?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestUpload.java
```diff
@@ -192,19 +204,23 @@ private void parseAndPersistFile(final UploadInformation the_info) {
...
-        final BallotManifestParser parser = new ColoradoBallotManifestParser(bmi_isr);
-        if (!parser.parse()) {
+        final BallotManifestParser parser = 
+            new ColoradoBallotManifestParser(bmi_isr, the_info.my_timestamp);
+        if (parser.parse()) {
+          Main.LOGGER.info(parser.parsedIDs().size() + 
+              " ballot manifest records parsed from upload file");
+          final OptionalLong count = count();
+          if (count.isPresent()) {
+            Main.LOGGER.info(count + " uploaded ballot manifest records in storage");
```
ranweiler: I think we now need `count.getAsLong()` here.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRUpload.java
dmzimmerman: It should indeed.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
dmzimmerman: I did this, but it will still need a bit of work later - right now it's still county names, but if we're going to send county ID numbers around instead (i.e., when you upload a file you specify the county that's uploading it by ID number, or we know because they're logged in what their ID is), that's an easy change later.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
kiniry: Which reminds me, we need to schedule to do a vulnerability scan, preferably before stage-3, though I know time is precious.  I recommend that we promise to CDOS to do it while they are doing user acceptance and load testing post-stage-3 delivery. (Please note, @sfsinger19103.)

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
```diff
@@ -237,6 +255,30 @@ public String endpoint(final Request the_request, final Response the_response) {
...
+    try {
+      Persistence.beginTransaction();
+      final Session s = Persistence.currentSession();
+      final Query<Long> query = 
+          s.createQuery("select count(1) from CastVoteRecord where record_type = '" + 
+                        RecordType.UPLOADED + "'", Long.class);
+      result = OptionalLong.of(query.getSingleResult());
+      Persistence.commitTransaction();
+    } catch (final PersistenceException e) {
+      // ignore
```
kiniry: Don't you mean `@todo dmz Figure out what to do here.`? ;)

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ContestDownloadByCounty.java
```diff
@@ -53,34 +54,68 @@ public EndpointType endpointType() {
...
-      return "";
-    } catch (final IOException e) {
-      the_response.status(HttpStatus.INTERNAL_SERVER_ERROR_500);
-      return "Unable to stream response.";
+      final Set<Contest> query_results = new HashSet<>();
+      Persistence.beginTransaction();
+      //this is very naive - it should be a single query that joins CVRs and contests
+      for (final String c : the_counties) {
+        final CastVoteRecord template =
+            new CastVoteRecord(RecordType.UPLOADED, null, c, null, null,
```
kiniry: I can't say I'm entirely comfortable with using normal constructors, rather than factories, to create templates vs. create actual instances.

server/eclipse-project/src/main/java/us/freeandfair/corla/hibernate/Persistence.java
```diff
@@ -40,6 +41,7 @@
 import us.freeandfair.corla.model.Contest;
 import us.freeandfair.corla.model.Elector;
 import us.freeandfair.corla.model.UploadedFile;
+import us.freeandfair.corla.util.Pair;
 
 /**
  * Manages persistence through Hibernate, and provides several utility methods.
```
kiniry: I feel like I need more of an explanation for the legitimate use of the helper methods in this class (someday).

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
```diff
@@ -224,6 +224,10 @@ public void start() {
     // provide properties to the persistence engine
     Persistence.setProperties(my_properties);
 
+    if (!Persistence.hasDB()) {
```
dmzimmerman: What is “the top-level”? If this particular error happens, the server doesn’t start.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownload.java
```diff
@@ -56,18 +62,50 @@ public String endpointName() {
...
       final OutputStream os = SparkHelper.getRaw(the_response).getOutputStream();
       final BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(os, "UTF-8"));
-      
-      Main.GSON.toJson(CastVoteRecord.getMatching(new HashSet<String>(), 
-                                                  RecordType.AUDITOR_ENTERED),
-                       bw);
-      bw.flush();
-      return "";
+      final Set<CastVoteRecord> matches = getMatching();
+      if (matches == null) {
```
dmzimmerman: Empty set, not null.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
dmzimmerman: I do need to fix the javadoc on that method, since it now returns an optional.

server/eclipse-project/src/main/java/us/freeandfair/corla/hibernate/Persistence.java
dmzimmerman: The quick explanation: if you're doing one query that's built in, just do it and the right thing will happen; if you're doing multiple operations, do:

`boolean transaction = beginTransaction() 
... operations ... 
if (transaction) { logic to commitTransaction() or rollbackTransaction() as appropriate }```

(with appropriate exception handling, of course)

This is essentially to work around the fact that Hibernate doesn't support nested transactions, but all the queries and object updates need to happen in transactions. So if you want to do something, but don't know if there's already a transaction happening, `beginTransaction()` tells you (and starts one if there wasn't one).

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownload.java
```diff
@@ -56,18 +62,50 @@ public String endpointName() {
...
+        status = HttpStatus.INTERNAL_SERVER_ERROR_500;
+      } else {
+        Main.GSON.toJson(matches, bw);
+        bw.flush();
+        result =  "";
+      }
     } catch (final IOException e) {
-      the_response.status(HttpStatus.INTERNAL_SERVER_ERROR_500);
-      return "Unable to stream response.";
+      status = HttpStatus.INTERNAL_SERVER_ERROR_500;
```
dmzimmerman: Agreed. I'll write up an issue for this one.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
dmzimmerman: As far as the persistence exception... well, the only thing we _can_ do there is return no results. It, like any other PersistenceException, is an unrecoverable exception (per Hibernate's docs).

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ContestDownloadByCounty.java
dmzimmerman: We should discuss this at some point. Honestly, if the thing had setter methods (which I'm considering just giving all of them anyway, as well as renaming methods to be more in line with JavaBeans conventions and having "get" prefixes), I'd construct one with default values and use setters for the template methods I'm using. If that's preferable, it's an easy thing to do. I don't see how having a factory would help.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -50,28 +54,67 @@ public EndpointType endpointType() {
    */
   @Override
   public String endpointName() {
-    return "/acvr/:counties";
+    return "/acvr/county";
```
ranweiler: That itemization will be very helpful.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -50,28 +54,67 @@ public EndpointType endpointType() {
...
       final BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(os, "UTF-8"));
-      
-      Main.GSON.toJson(CastVoteRecord.getMatching(county_set, 
-                                                  RecordType.AUDITOR_ENTERED),
-                       bw);
-      bw.flush();
-      return "";
+      final Set<CastVoteRecord> matches = getMatching(county_set);
+      if (matches == null) {
+        result = "Unable to fetch records from database";
```
ranweiler: We now have issue #186 in support of this.

PR #189, opened 2017-08-08
Improved queries and county ID persistence

dmzimmerman:
We needed a database table for counties (which will also incidentally be used in a future PR to track their contests); now we have one. Also, the queries on various endpoints now use type-safe JPA Criteria query syntax to (1) be more readable and type-checkable, and (2) protect against SQL injection attacks. Any queries left that are not converted yet will be converted in a later PR.


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestDownloadByCounty.java
```diff
@@ -67,28 +68,73 @@ public String endpointName() {
...
+      status = HttpStatus.BAD_REQUEST_400;
+      result = "invalid county ID specified";
     }
+    the_response.status(status);
+    return result;
+  }
+  
+  /**
+   * Validates the parameters of a request. For this endpoint, 
+   * the paramter names must all be integers.
```
ranweiler: `s/paramter/parameter`

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRDownloadByCounty.java
```diff
@@ -62,57 +69,98 @@ public String endpointName() {
...
+   * the paramter names must all be integers.
+   * 
+   * @param the_request The request.
+   * @return true if the parameters are valid, false otherwise.
+   */
+  private boolean validateParameters(final Request the_request) {
+    boolean result = true;
+    
+    for (final String s : the_request.queryParams()) {
+      try {
```
ranweiler: It would be nice to encapsulate this `try`/`catch` block in a function that just returns a `boolean`, which can then be used everywhere we are validating params.

server/eclipse-project/src/main/java/us/freeandfair/corla/gson/CVRContestInfoJsonAdapter.java
```diff
@@ -147,7 +149,7 @@ public CVRContestInfo read(final JsonReader the_reader)
           try {
             consensus = ConsensusValue.valueOf(the_reader.nextString());
           } catch (final IllegalArgumentException e) {
-            consensus = ConsensusValue.UNDEFINED;
+            // ignore
```
ranweiler: Maybe expand this comment to explain _why_ we are ignoring it.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyQueries.java
```diff
@@ -0,0 +1,138 @@
...
+      final CriteriaQuery<County> cq = cb.createQuery(County.class);
+      final Root<County> root = cq.from(County.class);
+      cq.select(root).where(cb.equal(root.get("my_identifier"), the_id));
+      final TypedQuery<County> query = s.createQuery(cq);
+      final List<County> query_results = query.getResultList();
+      // if there's exactly one result, return that
+      if (query_results.size() == 1) {
+        result = query_results.get(0);
+      } 
+      if (transaction) {
```
ranweiler: It would be really nice to encapsulate this transaction bookkeeping; it would be really easy to forget something. Maybe something that takes a lambda?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyQueries.java
```diff
@@ -0,0 +1,138 @@
...
+    } else {
+      Main.LOGGER.info("found county " + result);
+    }
+    return result;
+  }
+  
+  /**
+   * Obtains a County object from a county name or ID string. If the string is 
+   * numeric, we assume it is an ID; otherwise, we assume it is a name and match
+   * it as closely as we can to a county in the database.
```
ranweiler: I would prefer this be split up into different methods with smaller responsibilities. Any caller of this method should know whether its argument is meant to be an integer ID or a county name as string, right?

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
```diff
@@ -223,6 +234,44 @@ private void activateEndpoints() {
...
+    final Properties properties = new Properties();
+    try {
+      properties.load(ClassLoader.
+                      getSystemResourceAsStream(my_properties.getProperty(COUNTY_IDS)));
+    } catch (final IOException e) {
+      throw new IllegalStateException("Error loading county IDs, aborting.", e);
+    }
+    try {
+      for (final String s : properties.stringPropertyNames()) {
+        // if the property name is an integer, we assume it's a county ID
```
dmzimmerman: I don't really think so. That file is the source of truth about county IDs; there isn't really anywhere to look for other validation, and I'm not aware (in Java) of any way to check if something is an integer aside from trying to parse it as an integer.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
```diff
@@ -117,20 +119,27 @@ public ColoradoBallotManifestParser(final String the_string, final Instant the_t
...
-      return null;
-    } catch (final ArrayIndexOutOfBoundsException e) {
-      return null;
+      // obtain the county from the ID or name in the line
+      final County c = CountyQueries.fromString(the_line.get(COUNTY_ID_COLUMN));
+      if (c != null) {
+        result = 
+            Persistence.get(new BallotManifestInfo(the_timestamp, 
+                                                   c.identifier(),
+                                                   the_line.get(SCANNER_ID_COLUMN),
```
dmzimmerman: This is not about COUNTY_ID_COLUMN being null, it's about not finding a matching a county in the database. The others can't return null - the worst they'd do is say "there's no column with that number" (ArrayIndexOutOfBoundsException). Of course if the data in the file is bad, that's a problem, but there's no way for us to tell other than the number of ballots having to be numeric.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownload.java
```diff
@@ -87,23 +91,30 @@ public String endpoint(final Request the_request, final Response the_response) {
...
-          s.createQuery("from CastVoteRecord where record_type = '" + 
-              RecordType.AUDITOR_ENTERED + "'", CastVoteRecord.class);
-      query_result.addAll(query.getResultList());
-      Persistence.commitTransaction();
-      result = query_result;
+      final CriteriaBuilder cb = s.getCriteriaBuilder();
+      final CriteriaQuery<CastVoteRecord> cq = 
+          cb.createQuery(CastVoteRecord.class);
+      final Root<CastVoteRecord> root = cq.from(CastVoteRecord.class);
+      cq.select(root).where(cb.equal(root.get("my_record_type"),
```
dmzimmerman: `my_record_type` is the field of CastVoteRecord that corresponds to the database column that we're comparing for equality. There really isn't a way to construct queries without knowing either database column names or field names for the entities being turned into database rows.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownload.java
```diff
@@ -87,23 +91,30 @@ public String endpoint(final Request the_request, final Response the_response) {
...
+      final TypedQuery<CastVoteRecord> query = s.createQuery(cq);
+      result = new HashSet<CastVoteRecord>(query.getResultList());
+      try {
+        Persistence.commitTransaction();
+      } catch (final RollbackException e) {
+        Persistence.rollbackTransaction();
+      }
     } catch (final PersistenceException e) {
-      Main.LOGGER.error("Exception when reading ACVRs from database: " + e);
+      Main.LOGGER.error("Exception when reading ballot manifests from database: " + e);
```
dmzimmerman: Yup. Fixed in 65204f0

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -62,57 +69,98 @@ public String endpointName() {
...
-    try {
-      final OutputStream os = SparkHelper.getRaw(the_response).getOutputStream();
-      final BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(os, "UTF-8"));
+    if (validateParameters(the_request)) {
+      final Set<Integer> county_set = new HashSet<Integer>();
+      for (final String s : the_request.queryParams()) {
+        try {
+          county_set.add(Integer.valueOf(s));
+        } catch (final NumberFormatException e) {
+          // cannot happen because we validated the parameters
```
dmzimmerman: No, I probably just shouldn't catch that unchecked exception since it cannot happen. c17b61f

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -62,57 +69,98 @@ public String endpointName() {
...
+        }
+      }
       final Set<CastVoteRecord> matches = getMatching(county_set);
       if (matches == null) {
-        result = "Unable to fetch records from database";
         status = HttpStatus.INTERNAL_SERVER_ERROR_500;
+        result = "Error retrieving records from database";
```
dmzimmerman: Yes, that's right. Knowing what the exception was is immaterial because it, like all generic Hibernate persistence exceptions, is completely unrecoverable.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -62,57 +69,98 @@ public String endpointName() {
...
-        result =  "";
+        try {
+          final OutputStream os = SparkHelper.getRaw(the_response).getOutputStream();
+          final BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(os, "UTF-8"));
+
+          Main.GSON.toJson(matches, bw);
+          bw.flush();
+        } catch (final IOException e) {
+          status = HttpStatus.INTERNAL_SERVER_ERROR_500;
+          result = "Unable to stream response.";
```
dmzimmerman: Your call. I can go either way.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -62,57 +69,98 @@ public String endpointName() {
...
+        }
       }
-    } catch (final IOException e) {
-      status = HttpStatus.INTERNAL_SERVER_ERROR_500;
-      result = "Unable to stream response.";
+    } else {
+      status = HttpStatus.BAD_REQUEST_400;
```
dmzimmerman: If the requested county ID has no ACVRs, an empty list is returned; if the requested county ID is not a valid county ID, this situation happens. I don't know whether we should consider a request for a nonexistent county ID a bad request, or nonexistent data; I'm OK either way.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -62,57 +69,98 @@ public String endpointName() {
...
+      try {
+        Persistence.commitTransaction();
+      } catch (final RollbackException e) {
+        Persistence.rollbackTransaction();
       }
-      Persistence.commitTransaction();
-      result = query_result;
     } catch (final PersistenceException e) {
-      Main.LOGGER.error("Exception when reading ACVRs from database: " + e);
+      Main.LOGGER.error("Exception when reading ballot manifests from database: " + e);
```
dmzimmerman: 0b4383c

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRDownloadByCounty.java
dmzimmerman: Well, since we may be validating them different ways in different places, we might need many such functions; so far the validation has just been for integers but that may not end up being the case everywhere. I'll give it some more thought at some point.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestDownloadByCounty.java
dmzimmerman: 595db7c

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyQueries.java
dmzimmerman: Yes, I agree. I was actually thinking about just doing something like reference counting in the Persistence class's transaction tracking, and adding a method like "boolean commitOrRollbackTransaction()" for when you really just want to end the transaction and aren't going to do anything else interesting in the case of a rollback (which is most cases so far).

dmzimmerman: Of course, it is really important to get transaction tracking right - but they should be short-lived in the code, at least, such that you can _see_ the "begin" and "commit" matching each other. It's probably not possible to hide that complexity entirely.

server/eclipse-project/src/main/java/us/freeandfair/corla/gson/CVRContestInfoJsonAdapter.java
dmzimmerman: 93e39be

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyQueries.java
dmzimmerman: Unfortunately not. Because when we parse things like ballot manifest files, we literally have no (blank)ing clue what the data in that county field is going to look like. So we could do a "try it as an integer, then try it as a string"... but then we're doing that potentially in a few places.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
ranweiler: :+1:

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
ranweiler: No preference other than consistency.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
ranweiler: Let's go with a 404, since I see this situation as "we can't find one of the resources you identified in your URL", not "your request was malformed/too big/nonsensical".

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRDownloadByCounty.java
ranweiler: Sounds good.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyQueries.java
ranweiler: Oh wow, I didn't realize that those fields were so inconsistently-typed. This seems fine, then.

ranweiler:
Nice work, @dmzimmerman. I'd like to see that 400 become a 404, but otherwise this looks good to go.

PR #192, opened 2017-08-09
Add smoketest of cvr, manifest uploads to server

nealmcb:
At this point this is very messy, with some very very long lines of static data in auto-generated code,
no error handling in the main.py upload code, total kludge for dealing with ordering differences and ignoring timestamps (just do a fuzzy compare of the length of the response), etc.

But it runs a smoketest, testing several of the endpoints in corla-server-test.html,
and provides a basis for comment and refinement.


test/smoketest/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+The current tests are based on cvr at `test/e-1/arapahoe-regent-3-clear-CVR_Export.csv` and the manifest at `test/e-1/arapahoe-manifest.csv`.
+
+To run the smoketest, first cd to this directory.
+
+Start up a fresh server, or reset its database to be empty.
+
+Run `./main.py` to upload the cvr and manifest files.
+
+Run `pytest server_test.py` to test the
```
ranweiler: This sentence is accidentally truncated.

test/smoketest/main.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python
```
ranweiler: Are expecting Python 2 or 3?

test/smoketest/main.py
```diff
@@ -0,0 +1,31 @@
...
+    test_get_contest()
+    test_get_cvr_county_Arapahoe()
+    test_get_ballot_manifest_county_Arapahoe()
+    test_get_contest_id_2()
+    test_get_contest_county_Arapahoe()
+
+if __name__ == "__main__":
+    upload_cvr("../e-1/arapahoe-regent-3-clear-CVR_Export.csv")
+    upload_manifest("../e-1/arapahoe-manifest.csv")
+    # server_sequence()
```
ranweiler: Is this meant to be commented out?

nealmcb: Yes, for now. If I like zerotest and pytest, I assume I can figure out how to run the tests from main and update the README.md.

I've also resolved the other issues.
Both Python 2 and Python 3 work, but I'm mainly testing with Python 3.
I've made note of that.

nealmcb:
OK, this is updated to work on master 54a95d4, and be better documented and easier to update.
Please check it out and merge if you think it's useful.
Still lots to do to make it cover all the ASM transitions etc. which I'll continue in parallel or later.

dmzimmerman:
As far as I can see, the SQL query totals _all_ CVRs in the database, not just uploaded ones (or, more specifically, uploaded ones that are actually under audit - in the case of multiple uploads by a county to fix some sort of mistake, there will be many records that are not under audit, unless we purge them from the database... which we can, but currently don't, do). That might be fine if we're only running it once before any auditing happens and saving the results somewhere, though.

nealmcb:
Good point, @dmzimmerman. I think they will want a newly uploaded cvr or manifest to override any previous uploads, and not be included in these calculations. Offhand, for simplicity, I'd want all entries for that county removed from the database before adding the new one, and perhaps save the old file on the filesystem or something.
And @sfsinger19103, I think that at upload time, they will want the tabulation to be just for the subset uploaded by that county.
I think we do want to show results from all contests, since all contests are under audit. Only a few are used in the RLA calculations to set the sample size and calculate risk levels.

dmzimmerman:
I don’t think we ever want to remove an uploaded file from the database. I’m willing to be convinced that it makes sense to delete a county’s set of parsed CVRs when they upload new ones though, since they could always be re-parsed, and have the cast vote record table represent only the ones we’re actively working with. It occurs to me that uploading the file and parsing the CVRs is currently a very long lived operation (the database is locked for the entire duration of the file upload), and that may not make sense. @ranweiler, @kiniry, do you think it might make more sense to have an endpoint for uploading files where you specify the file type and hash (as one of the multipart/form-data parameters), get back a file ID of some sort, and subsequently tell the server “go ahead and make that my active file of that type”? (this same discussion applies to ballot manifest files too)

dmzimmerman:
Also, at the very least, I’d think this query would want to restrict itself to the set of CVRs for which the type is UPLOADED, because if we have to pick up where we left off at some point, we’d be out of luck unless we persistently save the tabulation results somewhere (which is do-able, of course).

ranweiler:
> do you think it might make more sense to have an endpoint for uploading files where you specify the file type and hash (as one of the multipart/form-data parameters), get back a file ID of some sort, and subsequently tell the server “go ahead and make that my active file of that type”?

@dmzimmerman, so do you mean decoupling the file upload from parsing? In other words, I submit `myCvrExport.csv` to `/file-upload` as form data with a `type` field of value `"cvrExport"`, then get an `id` back for that file. Then I would `POST` that `id` to another endpoint like `/parse-cvr`? Or do you mean something else?

kiniry:
Yes, I believe that is what @dmzimmerman is saying, @ranweiler.

test/smoketest/README.md
```diff
@@ -0,0 +1,147 @@
...
+This smoketest has been tested against master commit 54a95d4.
+
+For now, you'll need to install
+[zerotest](https://github.com/jjyr/zerotest)
+to run these tests, and/or generate new ones in the same way.
+
+This is mainly tested with python3, but has been seen to work
+on python2 also.
+
+`pip install zerotest`
```
kiniry: ... or 'pip3 install zerotest` (or similar) if you have multiple Python installations

test/smoketest/README.md
```diff
@@ -0,0 +1,147 @@
...
+ curl http://localhost:8888/contest/id/71
+ curl http://localhost:8888/cvr/county?3
+ curl http://localhost:8888/acvr/county?3
+ curl http://localhost:8888/acvr
+EndOfInput
+```
+
+Stop the `zerotest server` process, e.g. with Cntl-C.
+
+Interrim test of captured queries:
```
kiniry: s/Interrim/Interim/

test/smoketest/main.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
+"""Smoketest the RLA server
+TODO: get server_sequence working and displaying errors
```
kiniry: File an issue.

test/smoketest/smoketest-update.bash
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+set -eux -o pipefail
+
+# Put these somewhere else before running via travis
+export TRAVIS_BUILD_DIR=`git rev-parse --show-toplevel`
```
ranweiler: Is this meant to be run by Travis? If so, these variables should just be in the script's environment. I don't _think_ you should have to re-export them for the child scripts, either. Did that turn out to be necessary?

server/eclipse-project/pom.xml
```diff
@@ -4,9 +4,12 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>us.freeandfair.production</groupId>
 	<artifactId>colorado_rla</artifactId>
-	<version>0.0.1</version>
+	<version>0.7.0-alpha</version>
```
ranweiler: I don't think this PR is where we should be cutting releases.

test/smoketest/main.py
nealmcb: Filed: https://github.com/FreeAndFair/ColoradoRLA/issues/254

test/smoketest/smoketest-update.bash
nealmcb: I have not tried running this from Travis yet, but I intend to after figuring out what the python install pieces will look like etc.
For now, I'm just using the same environmental variables, and this was a quick hack to make them easy to set.

nealmcb: Putting the TEST directory in travis.yml just seems like a good hint to folks reviewing the travis setup that there is stuff close to being useful there. If you like I'll take it out, but it seems useful to me as is.

server/eclipse-project/pom.xml
nealmcb: That's fine. Just trying to keep the conversation going, and I wanted to test the release naming/launching process.
But this didn't seem like a particularly BAD PR to finally admit that we're nearing release, and can automatically test a bunch of stuff....
So I'll leave it in if I can get away with it :)

dmzimmerman: I agree with @ranweiler. Let's not do this here.

nealmcb: Makes sense.  I left in the fix for the build warnings.

We've been in 0.7.0-alpha for a week now, but I'll do it in a separate PR, after this PR lands.

test/smoketest/smoketest-update.bash
nealmcb: Removed, per feedback, in f3e381c

PR #196, opened 2017-08-09
Basic authentication for state/county admins

dmzimmerman:


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/Authentication.java
```diff
@@ -0,0 +1,157 @@
...
+  
+  /**
+   * Checks to see whether the specified request is authenticated as
+   * the specified administrator type.
+   * 
+   * @param the_request The request.
+   * @param the_type The administrator type.
+   * @return true if the request is authenticated, false otherwise.
+   */
+  public static boolean isAuthenticated(final Request the_request,
```
ranweiler: Maybe rename this to `isAuthenticatedAs()`, which would match the `authenticateAs()` method below.

dmzimmerman: Seems like a reasonable change.

dmzimmerman:
Yeah... I can't really do much more until we have some idea of how we're talking to two-factor authentication and the like. The basic bones of it in terms of how it's tracked within a session won't likely change much, though probably a sanity check that "this is a county administrator for the county they are trying to administer" will need to be added soon.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/Authentication.java
dmzimmerman: f3dd22d

ranweiler:
@dmzimmerman, the new commits look good! :ship:

PR #201, opened 2017-08-10
Fixing the random seed database storage issue

dmzimmerman:
Apparently, Hibernate chooses a dumb default when mapping `java.math.BigInteger`. I have shown it the error of its ways. I also enabled the real connection pool, which had been done in our default properties file but not passed on to Hibernate itself previously.

This is a continuation of the package-renaming branch (#200).


dmzimmerman:
(btw, whichever reviewer gets to this first, if you like it, just merge it; it's not necessary to have two other sets of eyes...)

PR #203, opened 2017-08-10
Adding the endpoint for selecting contests to audit.

dmzimmerman:
This also involves changing the data structures around a bit for the DoS dashboard, as I realized that storing a map of contests to reasons and a separate list of hand count contests makes less sense than just storing a bunch of "contest to audit" records with appropriate info in them.


dmzimmerman:
Note that this same endpoint can be used for `indicate-hand-count`, as the data structure it deals with and what it actually does with it are identical. If we want two endpoints that, respectively, ensure that one only deals with hand counts and one only with contests to comparison audit, we can do that straightforwardly.

server/eclipse-project/src/main/java/us/freeandfair/corla/json/ContestToAuditJsonAdapter.java
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * Checks the sanity of a contest to audit.
+   * 
+   * @param the_id The contest ID.
+   * @param the_reason The reason for audit.
+   * @param the_audit_type The audit type.
+   * @return the corresponding contest, if the data is sane, or null if
+   * the data is not.
+   */
+  private Contest sanityCheck(final Long the_id,
```
ranweiler: To me this method feels a bit inverted, and also makes more sense to have somewhere else. What is being checked has a domain meaning, not just serialization-level semantics.

Here's an idea: every `Contest` is a `PersistentEntity`. Would it be better to have `PersistentEntity` (and thus `Contest`) implement a `getByID` method, which returns either the entity, or `null`?

Then, the `Contest` class could implement `boolean isValidForAudit(AuditReason, AuditType)`, which takes an `AuditReason` and `AuditType` and performs this method's check, except for the `getByID` part. Or, if we had an `Audit` class, it could have a method which accepted a `Contest` and returned a `boolean`.

Then this:
```java
final Contest contest = sanityCheck(contest_id, reason, type);

if (error || contest == null) {
  throw new JsonSyntaxException("invalid data detected in contest to audit");
}
```
Would become:
```java
final Contest contest = Contest.getByID(contest_id);

if (error || contest == null || !contest.isValidForAudit(reason, type)) {
  throw new JsonSyntaxException("invalid data detected in contest to audit");
}
```

What do you think?

dmzimmerman: I've considered doing something like that... it definitely deserves some further thought. It might make more sense to break it into two different method calls (get the contest from Persistence, and then check the sanity of the other things), too. I don't know about adding a `getByID` to something like `PersistentEntity`, only because then the model classes are explicitly (albeit statically) talking to the persistence infrastructure; that's kind of the way it was before when I had that scattered across all those model classes.

ranweiler: > then the model classes are explicitly (albeit statically) talking to the persistence infrastructure; that's kind of the way it was before when I had that scattered across all those model classes.

Ah, I see. Because the models subclass from `PersistentEntity`, and have some persistence-related annotations, I didn't realize you were working to decouple them.

I'll let you mull over the big picture, but if you're trying to stick to a decoupled "data mapper" design, that seems good to me. I guess those validity checks that are dependent on persistence/the state of the world are an annoying edge case.

dmzimmerman:
Yeah... sadly, there’s no way around the annotations - but I view those as more saying “if you’re going to persist this entity, here’s how you do it” to the persistence framework, rather than as the entities implementing persistence related functionality themselves.

PR #210, opened 2017-08-11
Work on Endpoints, Queries and Dashboards

dmzimmerman:
Significant work on queries and endpoints: several endpoints that query for sets of data now stream those sets from the database, rather than pulling them all into memory at once (there is also a new streaming query API in `Persistence`). Also significant refinements to the county and audit board dashboard classes as part of making them persistent.


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRDownloadByCounty.java
```diff
@@ -77,21 +73,36 @@ public String endpoint(final Request the_request, final Response the_response) {
...
-        } catch (final IOException e) {
-          status = HttpStatus.INTERNAL_SERVER_ERROR_500;
-          result = "Unable to stream response";
+      try {
+        Persistence.beginTransaction();
+        final OutputStream os = SparkHelper.getRaw(the_response).getOutputStream();
+        final BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(os, "UTF-8"));
+        final JsonWriter jw = new JsonWriter(bw);
+        jw.beginArray();
+        for (final Integer county : county_set) {
```
ranweiler: It would be nice to encapsulate the serialization logic so that the endpoint logic proper were clearer. We'll want to do this when we add tests, anyway.

server/eclipse-project/src/main/java/us/freeandfair/corla/json/ContestToAuditJsonAdapter.java
```diff
@@ -72,6 +72,7 @@ public void write(final JsonWriter the_writer,
   }
   
   /**
+<<<<<<< HEAD
```
ranweiler: Stray merge conflict marker.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/AuditBoardDashboard.java
```diff
@@ -88,44 +133,57 @@ public void setMembers(final Collection<Elector> the_members) {
...
-      throw new IllegalArgumentException("attempt to set aCVR for unaudited CVR");
+  //@ require the_cvr_under_audit != null;
+  //@ require the_acvr != null;
+  public boolean submitAuditCVR(final CastVoteRecord the_cvr_under_audit, 
+                                final CastVoteRecord the_audit_cvr) {
+    // performs a sanity check to make sure the CVR under audit and the ACVR
+    // are the same card
+    boolean result = true;
+    final int index = my_cvrs_to_audit.indexOf(the_cvr_under_audit.id());
+    if (index > 0) {
```
ranweiler: Should this be `>= 0`, since it could be the first element in the array?

ranweiler: Also, suppose the ID doesn't appear in `my_cvrs_to_audit`. Then `index` is `-1`, so we return the default value of `result`, which is `true`. But we want to return `false` in that case, right?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/AuditBoardDashboard.java
```diff
@@ -88,44 +133,57 @@ public void setMembers(final Collection<Elector> the_members) {
...
+      }
     }
+    
+    return result;
   }
   
   /**
-   * @return all the aCVRs that have been submitted to this dashboard in 
-   * this round.
+   * @return
```
ranweiler: Is this comment now vestigial?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CastVoteRecord.java
```diff
@@ -257,6 +257,36 @@ public boolean equals(final Object the_other) {
...
+    if (the_other == null) {
+      result = false;
+    } else {
+      result &= nullableEquals(the_other.countyID(), countyID());
+      result &= nullableEquals(the_other.scannerID(), scannerID());
+      result &= nullableEquals(the_other.batchID(), batchID());
+      result &= nullableEquals(the_other.recordID(), recordID());
+      result &= nullableEquals(the_other.imprintedID(), imprintedID());
+      result &= nullableEquals(the_other.ballotType(), ballotType());
+      result &= recordType() == RecordType.AUDITOR_ENTERED ^
```
ranweiler: Ha, I just had a conversation the other day about the absence of an operator for logical xor in various languages (other than `!=`). It lives!

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
```diff
@@ -71,40 +96,39 @@ public County county() {
...
   }
   
   /**
-   * Sets new set of CVRs, replacing any current set of CVRs.
-   * 
-   * @param the_cvrs The CVRs.
+   * @return the CVR upload timestamp. A return value of null means
+   * that no CVRs have been uploaded for this county.
    */
-  public void setCVRs(final Collection<CastVoteRecord> the_cvrs) {
```
ranweiler: Was this unused?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
```diff
@@ -71,40 +96,39 @@ public County county() {
...
   }
   
   /**
-   * Sets the audit board membership.
+   * Sets a new CVR upload timestamp, replacing the previous one.
    * 
-   * @param the_board_members The members of the audit board.
+   * @param the_cvr_upload_timestamp The new upload timestamp.
    */
-  public void setAuditBoardMembership(final Collection<Elector> the_board_members) {
```
ranweiler: Did this go elsewhere?

server/eclipse-project/src/main/java/us/freeandfair/corla/query/CastVoteRecordQueries.java
```diff
@@ -0,0 +1,266 @@
...
+   * used within the same transaction.
+   *
+   * @param the_county The county.
+   * @param the_type The type.
+   * @return the stream of CastVoteRecord objects, or null if one could
+   * not be acquired.
+   * @exception IllegalStateException if this method is called outside a 
+   * transaction.
+   */
+  public static Stream<CastVoteRecord> getMatching(final Integer the_county,
```
ranweiler: With this and the above method, could we abstract out the stream creation and management to a method that e.g. takes a closure that accepts `cb` (the `CriteriaBuilder`) as an argument, then lets you build the query using that?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownload.java
```diff
@@ -72,51 +69,34 @@ public String endpoint(final Request the_request, final Response the_response) {
...
-      final TypedQuery<CastVoteRecord> query = s.createQuery(cq);
-      result = new HashSet<CastVoteRecord>(query.getResultList());
+      jw.beginArray();
+      final Stream<CastVoteRecord> matches = 
+          CastVoteRecordQueries.getMatching(RecordType.AUDITOR_ENTERED);
+      matches.forEach((the_cvr) -> {
+        try {
+          jw.jsonValue(Main.GSON.toJson(the_cvr));
+          Persistence.currentSession().evict(the_cvr);
+        } catch (final IOException e) {
```
dmzimmerman: Yes. But an IOException here means that our connection to the client is broken (as opposed to something going wrong with JSON serialization). So it probably doesn't matter.

dmzimmerman: We can certainly log it and break out of the for loop though.

dmzimmerman: Actually, we can't, because it's in a lambda. I think the only reason I had to catch the exception at all (instead of letting the `IOException` handler below deal with it) is because of that fact. Maybe I should save it and rethrow it after the `forEach()`...

server/eclipse-project/src/main/java/us/freeandfair/corla/model/AuditBoardDashboard.java
dmzimmerman: Addressed in 586e45f

server/eclipse-project/src/main/java/us/freeandfair/corla/model/AuditBoardDashboard.java
dmzimmerman: Yup.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CastVoteRecord.java
dmzimmerman: I know... It's so nice to get a chance to use it!

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
dmzimmerman: It's superfluous - can be obtained from a straightforward database query given the timestamp/countyID pair. So if we need to send the actual CVRs as part of the state we send the client, we'll build that state using this object and database queries.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
dmzimmerman: This is also superfluous, as above.

server/eclipse-project/src/main/java/us/freeandfair/corla/query/CastVoteRecordQueries.java
dmzimmerman: I'll have to think about that. I'm not used to thinking in streams and Java at the same time, this just seemed like a better idea than trying to use a JDBC `ScrollableResultSet` given its availability.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownload.java
ranweiler: That would be good. I just looked at the GSON source to verify. You're probably right about the underlying cause if we do throw an `IOException`, but I absolutely want to fail explicitly and fully if anything happens.

dmzimmerman: The current code does that in spectacular fashion. I'd never used `UncheckedIOException` before.

ranweiler: Done in 586e45fa8d40af360bd9f5809c4bdd77c3eb4a6c.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownloadByCounty.java
```diff
@@ -77,21 +73,36 @@ public String endpoint(final Request the_request, final Response the_response) {
...
+        final JsonWriter jw = new JsonWriter(bw);
+        jw.beginArray();
+        for (final Integer county : county_set) {
+          final Stream<CastVoteRecord> matches = 
+              CastVoteRecordQueries.getMatching(county, RecordType.AUDITOR_ENTERED);
+          matches.forEach((the_cvr) -> {
+            try {
+              jw.jsonValue(Main.GSON.toJson(the_cvr));
+              Persistence.currentSession().evict(the_cvr);
+            } catch (final IOException e) {
```
ranweiler: Done in 586e45fa8d40af360bd9f5809c4bdd77c3eb4a6c.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRDownload.java
```diff
@@ -72,15 +68,28 @@ public String endpoint(final Request the_request, final Response the_response) {
...
+      final JsonWriter jw = new JsonWriter(bw);
+      Persistence.beginTransaction();
+      jw.beginArray();
+      final Stream<CastVoteRecord> matches = 
+          CastVoteRecordQueries.getMatching(RecordType.UPLOADED);
+      matches.forEach((the_cvr) -> {
+        try {
+          jw.jsonValue(Main.GSON.toJson(the_cvr));
+          Persistence.currentSession().evict(the_cvr);
+        } catch (final IOException e) {
```
ranweiler: Done in 586e45fa8d40af360bd9f5809c4bdd77c3eb4a6c.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
ranweiler: :+1: just double-checking.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
ranweiler: :+1:

Issue #213, opened 2017-08-11
split authentication out into a separate ASM


kiniry:
Component ASMs in our specification relate to different UIs in the tool.  As such, the first authentication state and transitions should be in a trivial authentication ASM.

dmzimmerman:
This may not be quite as trivial as one might think, since authentication is likely going to be a multi-step process... and moreover, we'll probably need a new authentication ASM for every authentication that is attempted. I suppose we can do something like create the ASM with the first authentication step, return some kind of reference to it in the HTTP response, and have that sent with subsequent authentication steps... as well as some kind of garbage collection/timeout on old ASMs that don't get completed.

kiniry:
Has this been completed in #245, @dmzimmerman?

dmzimmerman:
It’s removed from the other ASMs. It does not have its own because right now it’s so simple it doesn’t really need one - a session is either authenticated or not, and can be authenticated to a different admin at any time. Which reminds me that we probably want to have an `unauthenticate` endpoint of some sort, though passing bad credentials to the `authenticate` ones has that effect now.

kiniry:
I think that design is appropriate for now, especially given our lack of knowledge about their two-factor authentication system as of now.  

I'm closing this issue for now and we can re-open it if/when appropriate.

I'll file a separate issue on the /unauthenticate endpoint.

PR #218, opened 2017-08-11
ASM precondition filter

kiniry:
This filter and a bunch of other code to simplify exceptional behavior handling are added through the introduction of a new base class for endpoints, `AbstractEndpoint`.


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ACVRDownload.java
```diff
@@ -66,28 +65,20 @@ public String endpointName() {
...
-    int status = HttpStatus.OK_200;
-    
     try {
       final OutputStream os = SparkHelper.getRaw(the_response).getOutputStream();
       final BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(os, "UTF-8"));
       final Set<CastVoteRecord> matches = getMatching();
       if (matches == null) {
-        result = "Unable to fetch records from database";
-        status = HttpStatus.INTERNAL_SERVER_ERROR_500;
+        serverError(the_response, "Unable to fetch records from database");
```
ranweiler: :+1:

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AuthenticateCountyAdministrator.java
```diff
@@ -54,15 +52,10 @@ public String endpointName() {
    */
   @Override
   public String endpoint(final Request the_request, final Response the_response) {
-    int status = HttpStatus.OK_200;
-    String result = "Authenticated";
-    
+    ok(the_response, "Authenticated");
```
ranweiler: Ah, these helpers _really_ improve readability.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
```diff
@@ -163,13 +165,13 @@ private void handleUpload(final Request the_request,
                 FileHelper.bufferedCopy(stream, os, BUFFER_SIZE, MAX_UPLOAD_SIZE);
 
             if (total >= MAX_UPLOAD_SIZE) {
-              Main.LOGGER.info("attempt to upload file greater than max size from " + 
+              Main.LOGGER.info("attempt to upload file greater than max size from " +
                                raw.getRemoteHost());
               the_info.my_response_string = "Upload Failed";
```
ranweiler: We should use a response helper here.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
```diff
@@ -231,10 +234,10 @@ private void parseAndPersistFile(final UploadInformation the_info) {
         the_info.my_ok = false;
         the_info.my_response_status = HttpStatus.UNPROCESSABLE_ENTITY_422;
```
ranweiler: We should use a response helper here.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
```diff
@@ -182,25 +184,25 @@ private void handleUpload(final Request the_request,
...
+
   /**
    * Parses an uploaded CVR export and attempts to persist it to the database.
    * 
    * @param the_info The upload information to use and update.
    */
   // the CSV parser can throw arbitrary runtime exceptions, which we must catch
   @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidCatchingNPE"})
-  private void parseAndPersistFile(final UploadInformation the_info) {
+  private void parseAndPersistFile(final Response the_response,
```
ranweiler: Hmm, it seems like `parseAndPersistFile()` shouldn't have to know anything about `the_response`.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
```diff
@@ -210,16 +212,17 @@ private void parseAndPersistFile(final UploadInformation the_info) {
...
         if (parser.parse()) {
-          Main.LOGGER.info(parser.parsedIDs().size() + " CVRs parsed from " + county + 
+          Main.LOGGER.info(parser.parsedIDs().size() + " CVRs parsed from " + county +
                            " county upload file");
           final OptionalLong count = count();
           if (count.isPresent()) {
             Main.LOGGER.info(count.getAsLong() + " uploaded CVRs in storage");
           }
-          attemptFilePersistence(the_info.my_file, county, hash, the_info.my_timestamp);
+          attemptFilePersistence(the_response, the_info.my_file, county, hash,
```
ranweiler: Maybe `attemptFilePersistence()` should just return a value that can be used, with `the_response`, in a helper like `serverError()` or similar.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ContestDownloadByID.java
```diff
@@ -60,15 +56,11 @@ public String endpoint(final Request the_request, final Response the_response) {
         result = Main.GSON.toJson(c);
       }
     } catch (final NumberFormatException e) {
-      status = HttpStatus.BAD_REQUEST_400;
-      result = "Bad contest ID";
+      invariantViolation(the_response, "Bad contest ID");
```
ranweiler: What status does this set? I assume `400 Bad Request`. Where can I find it?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
ranweiler: Yeah, in fact, I think that only `endpoint()` methods should ever call a helper like `serverError()`, `invariantViolation()`, &c.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestDownloadByCounty.java
```diff
@@ -88,16 +83,13 @@ public String endpoint(final Request the_request, final Response the_response) {
...
         } catch (final IOException e) {
-          status = HttpStatus.INTERNAL_SERVER_ERROR_500;
-          result = "Unable to stream response";
+          serverError(the_response, "Unable to stream response");
         }
       }
     } else {
-      status = HttpStatus.NOT_FOUND_404;
-      result = "Invalid county ID specified";
+      dataNotFound(the_response, "Invalid county ID specified");
```
ranweiler: This text could be improved. In other cases, "invalid" means "malformed" or "can't be parsed as the right type", and we call `invariantViolation()`. Here, the ID is well-formed, but doesn't have a representation in the database, so we (rightly) call `dataNotFound()`. Maybe we should change the other uses of "invalid" to "malformed"?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/UploadRandomSeed.java
```diff
@@ -86,26 +83,16 @@ public String endpoint(final Request the_request, final Response the_response) {
...
         Main.LOGGER.info("attempt to set an invalid random seed");
-        status = HttpStatus.BAD_REQUEST_400;
-        result = "Invalid random seed specified";
+        invariantViolation(the_response, "Invalid random seed specified");
       }
     } else {
       Main.LOGGER.info("unauthorized attempt to set the random seed");
-      status = HttpStatus.UNAUTHORIZED_401;
-      result = "Unauthorized attempt to set the random seed"; 
+      unauthorized(the_response, "Unauthorized attempt to set the random seed");
```
ranweiler: This will be pulled out into a filter in a later PR, correct?

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
     UPDATE_BALLOT_MARKS,
     UNDEFINED
   }
-%%% \todokiniry{Specifications for the transport types.}
+%%% We specify the content of data sent to/from the server either with
+%%% a specification of a URL parameter-based communication or using a
+%%% JSON object. We do not wish to precisely specify the JSON, as
+%%% encoding is non-unique due to the existenxce of semantically
```
ranweiler: s/existenxce/existence

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% equivalent encodings (as with any serialization technique) and
+%%% across libraries. Instead, we specify data-in-transit implicitly
+%%% via the function signatures in the theories above that focus on
+%%% the endpoint callback (the various \texttt{dashboard} theories)
+%%% and we name the semantially relevant fields using a list of
```
ranweiler: s/semantially/semantically

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
-    (: "status" :)
+%%% Legal values for \texttt{status} are: \texttt{too\_late},
+%%% \texttt{data\_\-transmission\_\-interrupted},
+%%% \texttt{upload\_\-sucessful}, \texttt{checking\_hash},
+%%% \texttt{hash\_wrong}, \texttt{hash\_verified},
+%%% \texttt{parsing\_\-data}, \texttt{file\_\-type\_\-wrong}, and
+%%% \texttt{data\_parsed}.
+    (: "ballot_manifest-status", "cvr-status" :)
+%%% There are several possible cases where a client can make an
+%%% inappropriate call to the server. Our client will be guarnateed to
```
ranweiler: s/guarnateed/guaranteed

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% never make sure a call, but we cannot presume that only our client
```
ranweiler: I can't parse this: "to never make sure a call"

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% will be communicating with the server, as the server will be
+%%% publicly accessible on the internet. We call these behaviors of
+%%% the server ``exceptional behavior''. All responses to exceptional
+%%% behavior come in the form of HTTP client error (\texttt{3xx}) or
```
ranweiler: You accidentally subtracted 100 from each of of the status code ranges: `3xx` status codes are for redirection; `4xx` are client errors, `5xx` are server.

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% server error (\texttt{4xx}) response codes.
```
ranweiler: As written, this definition claims that a `500` status code is in the range `4xx`.

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% can happen, e.g., when a client uses an improper encoding for a
+%%% piece of data, such as a county ID that does not exist.
+%%%
+%%% \item A piece of data uploaded by a client (at this point in time,
+%%% either a ballot manifest, a CVR file, or an Audit CVR) is
+%%% malformed and cannot be interpreted. This situation can happen,
+%%% for example, if the Election Management System used to export the
+%%% file has a bug, if the file is improperly edited by someone, or if
+%%% there is a bug in our parser. In this situation, the server will
+%%% respond with a \texttt{UNPROCESSABLE\_ENTITY\_422}
```
ranweiler: s/a/an

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% response.
+%%%
+%%% \item Alternatively, if the file is of the wrong type (i.e., a PDF
+%%% was uploaded when the server expected a CSV file), or if the
+%%% uploaded file is too large, then the server will respond with a
+%%% \texttt{UNSUPPORTED\_MEDIA\_TYPE\_415} message.
```
ranweiler: Will the server actually attempt to detect the content type of the file (which is sent as multipart form-encoded data)? I think in real life, this case will usually manifest as a failed file parse, hence a 422. We could simplify the spec by simply omitting this special case, and if a client sets an unsupported `content-type` header (e.g. `application/pdf`), just calling that a bad request or similar.

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%%
+%%% \item If the client is unauthenticated, or if authentication
+%%% fails, then the server will respond with a
+%%% \texttt{UNAUTHORIZED\_401} response.
+%%%
+%%% \item If some element of a request is incorrect then the server
+%%% responds with a \texttt{BAD\_REQUEST\_400} messages. In particular,
```
ranweiler: s/messages/message

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% if the digest uploaded with a ballot manifest or a CVR file is
+%%% incorrect, then the server will store the data and respond with
+%%% this message.
+%%%
+%%% \item Finally, if the endpoint called is improper given the state
```
ranweiler: Maybe clarify what "improper" means via e.g.: "if calling the given endpoint implies an illegal state transition,"

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% that the server is in, then the server will respond with a
+%%% \texttt{FORBIDDEN\_403} message. This happens only when an
```
ranweiler: I like this use of the 403 status.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
dmzimmerman: I agree, I'm not sure why it does.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
dmzimmerman: That sounds reasonable.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestDownloadByCounty.java
dmzimmerman: Maybe.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AuthenticateCountyAdministrator.java
dmzimmerman: Does this "ok" actually set the status, or send anything as a response? (it would help to be able to see the `ok` method, of course). We have to be careful with return values from `endpoint()`, because Spark will just send those out as part of the response; that's why, on the streaming ones, I return `""`.

specs/pvs/corla.pvs
dmzimmerman: Yes, I like it too.

dmzimmerman:
It will not surprise you, @kiniry, to learn that I like this refactoring a great deal.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -0,0 +1,194 @@
...
+
+    // This access is well-formed and permitted by the state machine, so log the
+    // use of the endpoint.
+    Main.LOGGER.info("checking authentication on " + the_request);
+
+    // If the client is unauthorized, then indicate such and redirect.
+
+    // If there is no authentication cookie in the request at all, then simply
+    // redirect to the top-level authentication endpoint.
+    // @todo kiniry Add once we have cookies turned on.
```
dmzimmerman: This functionality already works. But doing a redirect to the top-level authentication endpoint doesn't make a whole lot of sense in this context, to me, because the client has to display the authentication page itself - we're not displaying one.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -0,0 +1,194 @@
...
+
+    // If we have an authentication cookie, then check if the user is
+    // authenticated.
+    if (!Authentication.authenticate(the_request)) {
```
dmzimmerman: We need (later) to further get information from the state machine on what kind of authentication it needs (county or state) and check against that, probably.

specs/pvs/corla.pvs
dmzimmerman: s/sure/such

specs/pvs/corla.pvs
dmzimmerman: or just delete the "a"

specs/pvs/corla.pvs
dmzimmerman: I was, so far, not attempting to detect the content type. I didn't intend to do so. It's not even clear to me what the best way of doing so would be (beyond trying to parse it as CSV). Of course if people end up uploading Zip files full of JSON, well, that's a different story.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AuthenticateCountyAdministrator.java
dmzimmerman: I see that it does set the status, but I think that's OK as long as the status is correct at the end of the endpoint handler... it's not clear to me how Spark deals with it if one streams the response (by circumventing its cache), though, as we do in some cases.

specs/pvs/corla.pvs
```diff
@@ -2622,8 +2625,8 @@ BEGIN
 %%% administrators using the RLA Tool, cause state changes. These
 %%% events are named according to their corresponding functional
 %%% definitions in earlier theories. E.g., the
-%%% \texttt{authenticate\_state\_administrator\_event} relates to the
-%%% \texttt{authenticate\_state\_administrator} function and its
+%%% \texttt{authenticate\_\-state\_\-administrator\_event} relates to
+%%% the \texttt{authenticate\_state\_administrator} function and its
```
nealmcb: The definition in `county_dashboard_event` for `s_e` should not include states 1 or 9.

specs/pvs/corla.pvs
```diff
@@ -2800,23 +2803,112 @@ BEGIN
...
+%%% variety of response status the server can give to an upload. Note
+%%% that the \texttt{upload\_response} type is used as a response for
+%%% both ballot manifest and CVR uploads. These responses can happen
+%%% synchronously (our plan for \emph{stage-1} deliverable) or
+%%% asynchronously (our plan for a later deliverable). Asynchronous
+%%% updates happen via a call to the \texttt{refresh...} endpoints.
   upload_response: json_spec =
-    % legal values for status are:
-    % "too_late", "data_transmission_interrupted", "upload_sucessful",
-    % "checking_hash", "hash_wrong", "hash_verified", "parsing_data",
```
nealmcb: perhaps s/hash_wrong/hash_mismatch/ here and elsewhere, since we don't know if they got or entered the wrong value, or if there was file corruption along the way.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
kiniry: Done in eee59c0.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
kiniry: We need it to indicate which HTTP error codes should be sent to the client in the case of persistence failure.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
kiniry: I get the point, especially in instances where helper methods may choose to throw exceptions to indicate exceptional behavior.

kiniry: Would it be abusive to simply return HTTP error codes?  My problem is that this particular helper method actually has four exceptional behaviors...

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ContestDownloadByID.java
kiniry: Yes, it is a `400 Bad Request`.  What do you mean "Where can I find it?".  The implementation of `invariantViolation`?  It is in `AbstractEndpoint`.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotManifestDownloadByCounty.java
kiniry: I have not updated or addressed any of the terminology used in log messages.  We should (eventually) rewrite them to strictly use the terminology used in the documentation of the various exceptional behavior functions in `AbstractEndpoint`, which matches that which is documented in the specification.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/UploadRandomSeed.java
kiniry: Yes, when we figure out how authentication will actually work.  You can see the commented out logic about it in the `before` method.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
ranweiler: I think it is fine to be pragmatic and leave it as-is right now, since it is a private method.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ContestDownloadByID.java
ranweiler: You hadn't yet committed/pushed `AbstractEndpoint` when I wrote the above (-;

specs/pvs/corla.pvs
kiniry: I obviously didn't spell check or re-read this...

specs/pvs/corla.pvs
kiniry: Indeed, I'm not presuming we would do any sort of magic prefix analysis (though there are libraries for such), but I was just planning for this option.  Let me label this particular case as a potential future feature.

What about files that are too large though?  The HTTP spec says you should use a 415 in that case.

ranweiler: Actually, for that I guess 413 is most specific: https://tools.ietf.org/html/rfc7231#section-6.5.11.

I think a 400 would also be fine.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
kiniry: Agreed.  Let's do that on a new branch after my three relevant PRs are merged.

specs/pvs/corla.pvs
kiniry: I already filed a ticket on this.  The authentication state and transition should be properly extracted from these ASMs.  See #213.

specs/pvs/corla.pvs
kiniry: Sounds good.  Addressed in 808bb02.

dmzimmerman:
This PR should presumably be closed now, subsumed by #225

kiniry:
Closing this PR, now that all of its useful changes have been incorporated into `master`.

PR #220, opened 2017-08-12
ASM State Persistence

dmzimmerman:
This may not be quite the final model we go for, but as of now, the system persists 129 ASMs: one DoS dashboard ASM and 64 each of county and audit board ASMs (one for each county detected from the county list). It can restore these states at startup time, and if we use them correctly during endpoint computations, should be able to update them for every endpoint call.


server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
```diff
@@ -254,16 +274,62 @@ private void activateEndpoints() {
...
+   * Restores an ASM's state or persists it in the database.
    * 
-   * @exception IllegalStateException if we can't get the ASM. 
+   * @param the_asm The ASM.
+   * @param the_state The persistent state to restore, or null to persist
+   *  the state to the database.
+   * @exception PersistenceException if the state cannot be persisted.
    */
-  private void initializeASM() throws IllegalStateException {
+  private void restoreOrPersistState(final AbstractStateMachine the_asm,
```
ranweiler: I'm a little confused by the idea behind this method. It does opposite things, depending on if `the_state` is `null` or not, right? I guess it is private so it doesn't really matter.

ranweiler: Okay, I see this in context now and understand.

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
```diff
@@ -317,7 +388,7 @@ public void start() {
 
     if (Persistence.hasDB()) {
       initializeCounties();
-      initializeASM();
+      initializeASMs();
```
ranweiler: If this throws a `PersistenceError`, we're just letting it crash, right?

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
dmzimmerman: Yes. It was basically "I'm about to write the same code 3 times".

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
dmzimmerman: Yes. I don't see how we can possibly run without abstract state machines.

dmzimmerman: Or, rather, without _known valid_ ones. If we get a `PersistenceException` there, it means either "something has gone wrong and we can't talk to the database", or "I found inconsistent information in the database that makes me unable to determine the states of the ASMs" (such as two rows describing the same ASM). I could probably eliminate that possibility by making the primary key of that table be based on the class and identity of the ASM, but that's just pushing the problem elsewhere.

dmzimmerman:
BTW, specifically for @kiniry: "use them correctly" means that one must do the following:
- start a transaction
- get the appropriate ASM and do something to it
- get the appropriate persistent ASM state from Hibernate (with `PersistentASMStateQueries.get()`)
- call `updateFrom()` on that persistent ASM state, passing it the ASM
- commit the transaction

kiniry:
I have reviewed this PR and incorporated its design changes into the specification via the `asm-spec-updates-from-210-and-220` branch.  I'll open a PR on it now.

PR #223, opened 2017-08-12
Asm spec updates from 210 and 220

kiniry:


specs/pvs/corla.pvs
```diff
@@ -3152,6 +3193,12 @@ BEGIN
 %%% \todokiniry{Explain these relations and their use.}
   ui_to_asm_event_relation: pred[[ui_event, asm_event]]
   asm_event_to_endpoint_relation: pred[[asm_event, server_endpoint]]
+% %%% For convenience, sometimes it is useful to indicate to a state
+% %%% machine that it should transition based upon a given
+% %%% \texttt{ui\_event} (which, in turn, maps to a \texttt{asm\_event}
+% %%% via X).
```
ranweiler: What does `via X` mean?

kiniry: Heh.  A placeholder for me looking up the name of the relation.  I'll fix it.

kiniry: Addressed in b920a8a.

PR #226, opened 2017-08-12
Endpoint/ASM Work

dmzimmerman:
I started this branch to do work on integrating endpoints, ASMs, and authentication... but I'm not going to get very far tonight. Still, I made some notes and changes in `AbstractEndpoint` that ought not to be lost. @kiniry, feel free to pick up with this branch when you start working on this stuff in the morning.

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
@@ -26,9 +26,7 @@ import org.apache.commons.csv.CSVRecord;
 
 import us.freeandfair.corla.Main;
 import us.freeandfair.corla.model.BallotManifestInfo;
-import us.freeandfair.corla.model.County;
 import us.freeandfair.corla.persistence.Persistence;
-import us.freeandfair.corla.query.CountyQueries;
 
 /**
  * @description <description>
@@ -36,11 +34,13 @@ import us.freeandfair.corla.query.CountyQueries;
  * @bon OPTIONAL_BON_TYPENAME
  */
 public class ColoradoBallotManifestParser implements BallotManifestParser {
-  /**
-   * The column containing the county ID.
-   */
-  private static final int COUNTY_ID_COLUMN = 0;
-  
+//  TODO: if we want to validate county IDs against the county strings in
+//        the file, we'll need this later
+//  /**
+//   * The column containing the county ID.
+//   */
+//  private static final int COUNTY_ID_COLUMN = 0;
+ 
   /**
    * The column containing the scanner ID.
    */
@@ -87,27 +87,42 @@ public class ColoradoBallotManifestParser implements BallotManifestParser {
   private final Instant my_timestamp;
   
   /**
+   * The county ID to apply to the parsed manifest lines.
+   */
+  private final Integer my_county_id;
+  
+  /**
    * Construct a new Colorado ballot manifest parser using the specified Reader.
    * 
    * @param the_reader The reader from which to read the CSV to parse.
+   * @param the_timestamp The timestamp to apply to the parsed records.
+   * @param the_county_id The county ID for the parsed records.
    * @exception IOException if an error occurs while constructing the parser.
    */
-  public ColoradoBallotManifestParser(final Reader the_reader, final Instant the_timestamp) 
+  public ColoradoBallotManifestParser(final Reader the_reader, 
+                                      final Instant the_timestamp,
+                                      final Integer the_county_id) 
       throws IOException {
     my_parser = new CSVParser(the_reader, CSVFormat.DEFAULT);
     my_timestamp = the_timestamp;
+    my_county_id = the_county_id;
   }
   
   /**
    * Construct a new Colorado ballot manifest parser using the specified String.
    * 
    * @param the_string The CSV string to parse.
+   * @param the_timestamp The timestamp to apply to the parsed records.
+   * @param the_county_id The county ID for the parsed records.
    * @exception IOException if an error occurs while constructing the parser.
    */
-  public ColoradoBallotManifestParser(final String the_string, final Instant the_timestamp)
+  public ColoradoBallotManifestParser(final String the_string, 
+                                      final Instant the_timestamp,
+                                      final Integer the_county_id)
       throws IOException {
     my_parser = CSVParser.parse(the_string, CSVFormat.DEFAULT);
     my_timestamp = the_timestamp;
+    my_county_id = the_county_id;
   }
   
   /**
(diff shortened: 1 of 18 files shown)


dmzimmerman:
An important note on the rationale for the changes: transactions are per-thread. So whenever we start an endpoint, we should start a transaction; and whenever we exit an endpoint, we should either commit it or roll it back.

dmzimmerman:
We now have authorization support on endpoints (though not many of them take advantage of it yet), as well as transaction support so that simple query endpoints need not ever even worry about starting or ending transactions.

dmzimmerman:
Also, ballot manifest file uploads and CVR uploads now use the authenticated county administrator as the authoritative source of information about what county they belong to.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
```diff
@@ -122,19 +137,17 @@ private BallotManifestInfo extractBMI(final CSVRecord the_line,
...
-            Persistence.get(new BallotManifestInfo(the_timestamp, 
-                                                   c.identifier(),
-                                                   the_line.get(SCANNER_ID_COLUMN),
-                                                   the_line.get(BATCH_NUMBER_COLUMN),
-                                                   Integer.parseInt(the_line.
-                                                                    get(NUM_BALLOTS_COLUMN)),
-                                                   the_line.get(BATCH_LOCATION_COLUMN)),
-                            BallotManifestInfo.class);
-      }
+      // TODO: should we check for mismatched county IDs between the
```
kiniry: Wouldn't that be indicative of a pretty serious internal problem?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -83,6 +97,7 @@ public void badDataType(final Response the_response, final String the_log_messag
   public void unauthorized(final Response the_response, final String the_log_message) {
     the_response.status(HttpStatus.UNAUTHORIZED_401);
     Main.LOGGER.error("unauthorized access 401: " + the_log_message);
+    // TODO: should we halt() the endpoint execution here and simply send the response
```
kiniry: Probably, since authentication failures should reveal nothing about server state or behavior.  In fact, I was debating `halt`ing in all of these exceptional behavior methods, as such would simplify the logic of the various `endpoint()` methods.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -144,39 +159,76 @@ public void before(final Request the_request, final Response the_response) {
...
+    }
+  } 
+  
+  /**
+   * The after filter for this endpoint. By default, it attempts to commit 
+   * any open transaction (this makes writing endpoint code more straightforward,
+   * as the vast majority of endpoints will never have to deal with transactions
+   * themselves).
+   */
+  public void after(final Request the_request, final Response the_response) {
```
kiniry: Thumbs up on this pattern.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -144,39 +159,76 @@ public void before(final Request the_request, final Response the_response) {
...
+      } catch (final RollbackException e) {
+        // this is an internal server error because we don't know what didn't
+        // get committed
+        serverError(the_response, 
+                    "could not commit changes to persistent storage");
+        halt(the_response);
+        try {
+          Persistence.rollbackTransaction();
+        } catch (final PersistenceException ex) {
+          Main.LOGGER.error("could not roll back failed transaction: " + ex.getMessage());
```
kiniry: What the hell do we do then?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportUpload.java
```diff
@@ -99,32 +99,33 @@ public String endpointName() {
...
                                 HashStatus.NOT_CHECKED, blob);
       Persistence.saveOrUpdate(result);
-      if (transaction) {
-        try {
-          Persistence.commitTransaction();
-        } catch (final RollbackException e) {
-          Persistence.rollbackTransaction();
-        }
-      }
+      // TODO: currently we have to commit this transaction here, because
```
kiniry: Let's file an issue on that!

server/eclipse-project/src/main/java/us/freeandfair/corla/model/AuditBoardDashboard.java
```diff
@@ -179,4 +192,14 @@ public boolean submitAuditCVR(final CastVoteRecord the_cvr_under_audit,
...
   }
+  
+  /**
+   * @return the list of audit CVRs submitted. The result will contain
+   * a null value for each element in the cvrsToAudit() list where there
+   * has been no audit CVR submitted. Thus, most computations on this list
+   * will use the sublist preceding the first null value.
+   */
+  public List<Long> submittedAuditCVRs() {
+    return Collections.unmodifiableList(my_submitted_audit_cvrs);
```
kiniry: Nice.  I learned something new.  I haven't been using immutability like this on references that have been leaking through APIs.  We should address that in one go later one.  Another issue to get filed.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
dmzimmerman: It would, but for Dominion, not for us. It would mean that a county was uploading ballot manifests with the wrong county name in them.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
dmzimmerman: The problem is, you can only `halt()` in the `before` part... not during actual endpoint execution. You can see how I addressed that in the upload endpoints.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
dmzimmerman: Well... it really should never happen; it means the database is so hopelessly broken that it won't let you roll back - or it means the database went away, in which case we fail over and try again.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/AuditBoardDashboard.java
dmzimmerman: I've been trying to do it pretty consistently.

PR #235, opened 2017-08-12
Asm load-transition check-transition-save support

kiniry:
This branch implements automatic ASM load, ASM state transition legality, transition, and save via `before` and `after` actions.  Look at `AbstractEndpoint` and `AbstractDoSEndpoint` to perform a design review.

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/asm/AbstractStateMachine.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/asm/AbstractStateMachine.java
@@ -12,6 +12,10 @@
 
 package us.freeandfair.corla.asm;
 
+import static us.freeandfair.corla.asm.ASMEvent.AuditBoardDashboardEvent.*;
+import static us.freeandfair.corla.asm.ASMEvent.CountyDashboardEvent.*;
+import static us.freeandfair.corla.asm.ASMEvent.DoSDashboardEvent.*;
+
 import java.io.Serializable;
 import java.util.HashSet;
 import java.util.List;
@@ -176,6 +180,13 @@ public abstract class AbstractStateMachine implements Serializable {
    */
   public Set<ASMEvent> enabledASMEvents() {
     final Set<ASMEvent> result = new HashSet<ASMEvent>();
+    // skips and refreshes are always permitted
+    result.add(DOS_SKIP_EVENT);
+    result.add(DOS_REFRESH_EVENT);
+    result.add(COUNTY_SKIP_EVENT);
+    result.add(COUNTY_REFRESH_EVENT);
+    result.add(AUDIT_SKIP_EVENT);
+    result.add(AUDIT_REFRESH_EVENT);
     for (final ASMTransition t : my_transition_function) {
       if (t.startState().equals(my_current_state)) {
         result.add(t.event());
@@ -217,6 +228,16 @@ public abstract class AbstractStateMachine implements Serializable {
    */
   public ASMState stepEvent(final ASMEvent the_event)
       throws IllegalStateException {
+    // refresh and skip events are always permitted
+    if (the_event == DOS_REFRESH_EVENT ||
+        the_event == DOS_SKIP_EVENT ||
+        the_event == COUNTY_REFRESH_EVENT || 
+        the_event == COUNTY_SKIP_EVENT ||
+        the_event == AUDIT_REFRESH_EVENT || 
+        the_event == AUDIT_SKIP_EVENT) {
+      return currentState();
+    }
+                
     ASMState result = null;
     for (final ASMTransition t : my_transition_function) {
       if (t.startState().equals(my_current_state) &&
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/asm/AbstractStateMachine.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/asm/AbstractStateMachine.java
@@ -30,6 +30,8 @@ import us.freeandfair.corla.Main;
  * @author Daniel M. Zimmerman <dmz@freeandfair.us>
  * @version 0.0.1
  */
+// we'll need to investigate the complexity of this class later
+@SuppressWarnings("PMD.CyclomaticComplexity")
 public abstract class AbstractStateMachine implements Serializable {
   /**
    * The serialVersionUID.
@@ -226,6 +228,7 @@ public abstract class AbstractStateMachine implements Serializable {
    * @throws IllegalStateException is this ASM cannot transition given
    * the provided event.
    */
+  @SuppressWarnings("PMD.CyclomaticComplexity")
   public ASMState stepEvent(final ASMEvent the_event)
       throws IllegalStateException {
     // refresh and skip events are always permitted
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractDoSEndpoint.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractDoSEndpoint.java
@@ -1,0 +1,36 @@
+/*
+ * Free & Fair Colorado RLA System
+ * 
+ * @title ColoradoRLA
+ * @created Aug 12, 2017
+ * @copyright 2017 Free & Fair
+ * @license GNU General Public License 3.0
+ * @creator Joe Kiniry <kiniry@freeandfair.us>
+ * @description A system to assist in conducting statewide risk-limiting audits.
+ */
+
+package us.freeandfair.corla.endpoint;
+
+import us.freeandfair.corla.asm.DoSDashboardASM;
+
+/**
+ * Functionality that spans endpoints on the Department of State Dashboard.
+ */
+@SuppressWarnings("PMD.AtLeastOneConstructor")
+public abstract class AbstractDoSEndpoint extends AbstractEndpoint {
+  /**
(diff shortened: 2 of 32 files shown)


kiniry:
I will now implement the other two new classes following this design while @dmzimmerman reviews.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -39,50 +50,78 @@
...
+  protected abstract Class<? extends AbstractStateMachine> asmClass(); 
+  
+  /**
+   * Load the appropriate ASM from the database and make sure that
+   * the transition we wish to take is legal.
+   */
+  protected void loadAndCheckASM(final Response the_response) {
+    // get the state of the DoS dashboard
+    try {
+      my_asm = (AbstractStateMachine) (asmClass().getConstructor().newInstance());
```
dmzimmerman: This will not work for the ASMs that have county numbers.

dmzimmerman: (neither will the query below)

BTW, if we're going to just construct a new one every endpoint, I guess there's no real need to store them in Main. Which may be something you also changed, but I haven't gotten to the end yet.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -39,50 +50,78 @@
...
-   * Indicate that data that the client provided is ill-formed.
+   * Save the ASM back to the database.
    */
-  public void badDataContents(final Response the_response, final String the_log_message) {
-    the_response.status(HttpStatus.UNPROCESSABLE_ENTITY_422);
-    Main.LOGGER.error("bad data from client 422: " + the_log_message);
-    my_endpoint_result = the_log_message;
+  protected void transitionAndSaveASM()  {
+    my_asm.stepEvent(endpointEvent());
+    final PersistentASMState new_state = PersistentASMState.stateFor(my_asm);
```
dmzimmerman: This needs to take the actual persistent state object (either because we save it somewhere, or because we ask for it again) returned by `PersistentASMStateQueries`and call `updateFrom()` on it. This new object will have a different database key, and when you save it, there will then be two different states for the same ASM (and the system will understandably get upset next time it tries to load one).

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractDoSEndpoint.java
```diff
@@ -0,0 +1,36 @@
...
+    return AuthorizationType.STATE;
+  }
+  
+  /**
+   * @return subclasses of this endpoint use the Department of State ASM.
+   */
+  @Override
+  protected Class<DoSDashboardASM> asmClass() {
+    return DoSDashboardASM.class;
+  }
```
dmzimmerman: I think we need another method here, something like `protected String asmIdentity(final Request the_request)` that returns the ASM's identity string; for the DoS dashboard it'd return `null`, because that's a singleton (we could also choose something else consistent other than `null`), but for the others, it would check the request and extract the county ID that the county administrator is authenticated as (using `Authentication.authenticatedCounty(Request)`).

There may be other ways to do this too.

server/eclipse-project/src/main/java/us/freeandfair/corla/asm/AbstractStateMachine.java
```diff
@@ -217,6 +228,16 @@ public ASMState stepTransition(final ASMTransition the_transition)
    */
   public ASMState stepEvent(final ASMEvent the_event)
       throws IllegalStateException {
+    // refresh and skip events are always permitted
```
dmzimmerman: Eek. I don't like having these in the AbstractStateMachine class directly. Is there a way that we can annotate the events in the enumeration instead (like a method `isAlwaysEnabled()`), or throw them in a collection of always-enabled events that can be added to by child classes, or something?

dmzimmerman: (the same goes for their occurrence above)

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -47,57 +52,83 @@
...
+  public void ok(final Response the_response) {
+    the_response.status(HttpStatus.OK_200);    
   }
-
+  
   /**
-   * Indicate that the client is not authorized to perform the requested action.
+   * Indicate and log that the operation completed successfully.
    */
+<<<<<<< HEAD
```
dmzimmerman: Some merge remnants here...

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -47,57 +52,83 @@
...
   public void unauthorized(final Response the_response, final String the_log_message) {
     the_response.status(HttpStatus.UNAUTHORIZED_401);
     Main.LOGGER.error("unauthorized access 401: " + the_log_message);
     // TODO: should we halt() the endpoint execution here and simply send the response
+=======
```
dmzimmerman: And more merge remnants.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -47,57 +52,83 @@
...
+  public void ok(final Response the_response, final String the_log_message) {
+    the_response.status(HttpStatus.OK_200);
+    Main.LOGGER.error("successful operation 200: " + the_log_message);
+>>>>>>> Implemented automatic ASM load, transition, and save for endpoints.
```
dmzimmerman: And even more merge remnants.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -202,6 +262,12 @@ public void before(final Request the_request, final Response the_response) {
    * themselves).
    */
   public void after(final Request the_request, final Response the_response) {
+<<<<<<< HEAD
```
dmzimmerman: I'm going to stop marking merge remnants.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ContestDownloadByID.java
```diff
@@ -25,7 +28,7 @@
  * @version 0.0.1
  */
 @SuppressWarnings("PMD.AtLeastOneConstructor")
-public class ContestDownloadByID extends AbstractEndpoint {
+public class ContestDownloadByID extends AbstractCountyDashboardEndpoint {
```
dmzimmerman: This endpoint might also need to be accessed by the state dashboard to get contest information for display.

server/eclipse-project/src/main/java/us/freeandfair/corla/asm/AbstractStateMachine.java
kiniry: I don't like it either, but we gotta get this done.  I'll file a ticket for a refactor.

kiniry: Filed in issue #239.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
kiniry: Yes, they are taken care of now.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ContestDownloadByID.java
kiniry: Indeed.  There were a couple of these that I couldn't tell which dashboard they were for.  Given that they are internal private skip events, it doesn't really matter.  In fact, maybe this kind of situation should just inherit from `AbstractEndpoint`?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AuthenticateCountyAdministrator.java
```diff
@@ -25,6 +29,22 @@
...
   /**
+   * @return no authorization is required for this endpoints.
+   */
+  @Override
+  public AuthorizationType requiredAuthorization() {
+    return AuthorizationType.NONE;
+  }
+  
+  /**
+   * @return this endpoint uses the Department of State ASM.
```
dmzimmerman: Does it? (I assume we can easily fix this comment, the code looks fine)

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/ContestDownloadByID.java
dmzimmerman: Indeed. I suspect that such events shouldn't even touch state machines.

dmzimmerman:
This was subsumed by #242 as well.

PR #237, opened 2017-08-13
Adding support for phantom ballots

dmzimmerman:
Seeing the `ballot_not_found` endpoint spec made me realize that we needed a fourth CVR type.


dmzimmerman:
This is work in progress in the sense that the endpoint isn't ASM-savvy, but otherwise it should work fine (to the extent that I am able to test it without being able to start an audit).

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/BallotNotFound.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public String endpointName() {
+    return "/ballot-not-found";
+  }
+
```
kiniry: You forgot
```
  /**
   * @return COUNTY authorization is required for this endpoint.
   */
  public AuthorizationType requiredAuthorization() {
    return AuthorizationType.COUNTY;
  }
```

kiniry: This needs to be fixed so that'll it'll merge with the other branch.

dmzimmerman:
I am moving the changes from here to #245, and deleting this branch.

PR #240, opened 2017-08-13
Missing endpoints

kiniry:
Here are initial/template endpoint implementations for all missing inbound public event endpoints.

PR #242, opened 2017-08-13
Endpoint asm merge

kiniry:
This branch contains all work in the `missing-endpoints` branch covered by PR #240 and the `asm-load-transition-save` branch covered in PR #235.  It is ready for review, merge into master, and debugging.

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/IndicateHandCount.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/IndicateHandCount.java
@@ -12,18 +12,12 @@
 
 package us.freeandfair.corla.endpoint;
 
-import javax.persistence.PersistenceException;
-
-import com.google.gson.JsonSyntaxException;
+import static us.freeandfair.corla.asm.ASMEvent.DoSDashboardEvent.INDICATE_FULL_HAND_COUNT_CONTEST_EVENT;
 
 import spark.Request;
 import spark.Response;
 
-import us.freeandfair.corla.Main;
-import us.freeandfair.corla.model.ContestToAudit;
-import us.freeandfair.corla.model.DepartmentOfStateDashboard;
-import us.freeandfair.corla.persistence.Persistence;
-import us.freeandfair.corla.query.DepartmentOfStateDashboardQueries;
+import us.freeandfair.corla.asm.ASMEvent;
 
 /**
  * The endpoint for indicating that a contest must be hand-counted.
@@ -77,7 +71,8 @@ public class IndicateHandCount extends AbstractDoSDashboardEndpoint {
     // try {
     //   final ContestToAudit[] contests = 
     //       Main.GSON.fromJson(the_request.body(), ContestToAudit[].class);
-    //   final DepartmentOfStateDashboard dosdb = DepartmentOfStateDashboardQueries.get();
+    //   final DepartmentOfStateDashboard dosdb = 
+    //       DepartmentOfStateDashboardQueries.get();
     //   if (dosdb == null) {
     //     Main.LOGGER.error("could not get department of state dashboard");
     //     serverError(the_response, "Could not select contests");
@@ -96,6 +91,6 @@ public class IndicateHandCount extends AbstractDoSDashboardEndpoint {
     //   serverError(the_response, "Unable to save contest selection");
     // }
     // return my_endpoint_result;
-    return "Selected the passed contests for hand audit."
+    return "Selected the passed contests for hand audit.";
   }
 }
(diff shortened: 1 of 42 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/IndicateHandCount.java
```diff
@@ -0,0 +1,96 @@
...
+   * Indicate that a contest must be hand-counted.
+   * 
+   * @param the_request The request.
+   * @param the_response The response.
+   */
+  @Override
+  public synchronized String endpoint(final Request the_request, 
+                                      final Response the_response) {
+    ok(the_response, "Contests selected");
+    // try {
```
dmzimmerman: This is really going to be exactly the same code, unless we take the trouble to come up with another data format. What will likely change is some amount of data validation where we'll check to see that the ContestToAudit objects have the right audit type in them for the particular endpoint.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/IntermediateAuditReport.java
```diff
@@ -0,0 +1,71 @@
...
+    return SUBMIT_INTERMEDIATE_AUDIT_REPORT_EVENT;
+  }
+  
+  /**
+   * Publish the intermediate audit report by the audit board.
+   */
+  @Override
+  public String endpoint(final Request the_request,
+                         final Response the_response) {
+    return "Save an intermediate audit report so that the audit board members " +
```
dmzimmerman: I don't understand this endpoint. What is an "intermediate audit report", and how does it differ from the audit state that we always keep on the server after every action?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/PublishBallotsToAudit.java
```diff
@@ -0,0 +1,68 @@
...
+import us.freeandfair.corla.asm.ASMEvent;
+
+/**
+ * Download all ballots to audit for the entire state.
+ * 
+ * @author Joe Kiniry <kiniry@freeandfair.us>
+ * @version 0.0.1
+ */
+@SuppressWarnings("PMD.AtLeastOneConstructor")
+public class PublishBallotsToAudit extends AbstractDoSDashboardEndpoint {
```
dmzimmerman: Does "Publish" mean "download" as in "you contact this endpoint and it gives you the list of ballots", or does it mean "we set up the internal data structures so that the various dashboard refresh endpoints will have the lists of ballots", or does it mean something else?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/IndicateHandCount.java
kiniry: Yes, I agree.  We could just do it with a parameter.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/IntermediateAuditReport.java
kiniry: It is in the spec, as you suggested.  CDOS asked us to have the audit board submit an intermediate report when they want to take a break, go to lunch, go home for the night, etc.  The report itself is just a small textfield where they can type in the reason that they are taking a break and anything else that might be salient. (CC @ranweiler)

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/PublishBallotsToAudit.java
kiniry: I had taken it to mean the former, as it is a GET with a JSON result.  If you think it wiser to just make it a trigger, that's fine too.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/IntermediateAuditReport.java
dmzimmerman: Ah, OK. The "small text field" part was not clear to me. One presumes, though, that they would need to be able to do this more than once? Do we save all of them? Timestamp them? (I sense more database schema changes on the horizon...)

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/PublishBallotsToAudit.java
dmzimmerman: I guess my question is, who uses the result of this endpoint, if it is indeed just a list of all the ballots to audit statewide?

PR #243, opened 2017-08-13
Hash checking

kiniry:
Here is a spec and implementation of SHA-256.

Note that we will have to make the comparison of the hex-encoded digest robust to silly things like county administrators changing the case of the `X`.


specs/pvs/corla.pvs
```diff
@@ -1088,7 +1092,7 @@ BEGIN
 %%% quoted below.
 
 % @trace RDPR-6-Jul-2017 Section 25.2.2(H)
-
+e
```
nealmcb: Looks like you don't want this....

specs/pvs/corla.pvs
```diff
@@ -1118,7 +1122,7 @@ BEGIN
 %%% As such, the seed of our pseudorandom algorithm must be a natural
 %%% number that is at least twenty digits long.
   seed: TYPE = {n: nat | 9999999999999999999 < n}
-%%% The random number generator itself must be seeded with the seed
+s%%% The random number generator itself must be seeded with the seed
```
nealmcb: typo

kiniry:
I have corrected the typos, rebuilt the specification book, and checked that the spec is type-safe.

I agree.  Java is incredibly verbose.  This is why, e.g., our sample implementation of STAR-Vote, is 1/9th the size of the example Java implementation from Rice.  When the opportunity presents itself, we definitely choose to use more modern languages than Java (or C, or C++, or ...).

PR #245, opened 2017-08-13
More Endpoint/ASM Work

dmzimmerman:
A lot of work on filling in endpoints and getting ASMs to work correctly in the presence of authentication.


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -55,6 +55,11 @@
 @SuppressWarnings({"PMD.AtLeastOneConstructor", "PMD.TooManyMethods"})
 public abstract class AbstractEndpoint implements Endpoint {
   /**
+   * A flag that disables ASM checks, when true.
```
kiniry: Great.  Do we also need a `DISABLE_AUTHENTICATION` for `phase-1`?  See #248.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/DoSDashboardASMState.java
```diff
@@ -0,0 +1,68 @@
...
+  public String endpointName() {
+    return "/dos-asm-state";
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public String endpoint(final Request the_request, final Response the_response) {
+    // there's really nothing to do here other than get the ASM state, which we
```
kiniry: Right on!

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
kiniry: Also, we should probably make this a property.  See #252.

PR #257, opened 2017-08-13
The audit board upload endpoint.

dmzimmerman:
Also fixed a minor problem with authentication when no request body is submitted.


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/EstablishAuditBoard.java
```diff
@@ -62,8 +77,40 @@ protected ASMEvent endpointEvent() {
...
+  // false positive about inner class declaration
+  @SuppressFBWarnings("SIC_INNER_SHOULD_BE_STATIC_ANON")
   public String endpoint(final Request the_request,
                          final Response the_response) {
-    return "The audit board specified has been saved.";
+    try {
+      final Type list_type = new TypeToken<List<Elector>>() { }.getType();
+      final List<Elector> audit_board = 
+          Main.GSON.fromJson(the_request.body(), list_type);
+      if (audit_board.size() >= CountyDashboard.MIN_AUDIT_BOARD_MEMBERS) {
```
kiniry: I think that this is == in the spec, but I'm ok with leaving it this way for now.

PR #261, opened 2017-08-13
First example auto-generated JSON schemas.

kiniry:
The web server https://jsonschema.net/#/editor was used to process
captured JSON samples from Neal's smoketest work.

### DIFF
--- a/docs/json/acvr.json
+++ b/docs/json/acvr.json
@@ -1,0 +1,76 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "definitions": {},
+    "id": "http://example.com/example.json",
+    "properties": {
+        "ballot_type": {
+            "id": "/properties/ballot_type",
+            "type": "string"
+        },
+        "batch_id": {
+            "id": "/properties/batch_id",
+            "type": "string"
+        },
+        "contest_info": {
+            "id": "/properties/contest_info",
+            "items": {
+                "id": "/properties/contest_info/items",
+                "properties": {
+                    "choices": {
+                        "id": "/properties/contest_info/items/properties/choices",
+                        "items": {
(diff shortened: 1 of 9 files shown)


kiniry:
How does this kind of thing look, @ranweiler?

ranweiler (approved):
@kiniry, this is great, and will be extremely helpful. One question: did you mean to commit `server_test-pp.json`?

dmzimmerman:
Those came out better than I would have expected. I like it, but I also think `server_test-pp.json` was probably an unintentional commit.

dmzimmerman (approved):

ranweiler:
I'd love it if we added such schemas for all endpoints, and merged this in.

kiniry:
We'll have to complete this PR in the coming week or two via the ongoing work in the `json-schemas` branch, particularly using the idea embodied in 3cbadd3.

dmzimmerman:
Indeed. I would argue that we should actually rename that particular attribute, too.

nealmcb:
I'd support renaming it. Easy to change it in main.py.  I was going to suggest that, but didn't know how much work would be involved.

nealmcb:
After exploring this some more in the context of #845, Something more like genson (the powerful, user-friendly JSON Schema generator in Python: https://pypi.python.org/pypi/genson/ not Java genson) would seem more helpful. It makes it easier to document evolving schemas, since you can feed it as many schemas and objects as you want and it will spit out one, unified schema for them all.
In particular I think we want to be able to manually edit descriptions of the elements, especially the ones that are complex and important to describe, but still update it all in an automatable way when a new element is added to a schema, etc.

I found several of the answers and comments at [Tool to generate JSON schema from JSON data - Stack Overflow](https://stackoverflow.com/questions/7341537/tool-to-generate-json-schema-from-json-data) to be useful.

PR #263, opened 2017-08-13
Audit board dashboard and friends

kiniry:


server/eclipse-project/src/main/java/us/freeandfair/corla/json/CountyDashboardRefreshResponse.java
```diff
@@ -192,6 +212,8 @@ protected CountyDashboardRefreshResponse(final CountyStatus the_status,
     final UploadedFile cvr_file = 
         UploadedFileQueries.matching(county_id, the_dashboard.cvrUploadTimestamp(), 
                                      FileType.CAST_VOTE_RECORD_EXPORT);
+    // @todo kiniry This should have been/will be saved in the DB at the time
+    // in which the CVR file was uploaded.
```
dmzimmerman: It already is, assuming that we won't use a file where the hash doesn't match. The hash checking at upload time will be added soon, once your hash class is finished.

PR #273, opened 2017-08-15
final pass on updating documentation for the stage-1 deliverable

kiniry:
E.g., my first comment provides a new introduction to the spec, given the adoption of CCR 1505-1.


specs/pvs/corla.pvs
```diff
@@ -930,6 +947,7 @@ BEGIN
        election: election,
        audit_board: audit_board,
        county_administrator: county_administrator #]
+% @review CDOS What must be included according to (E) above?
```
sfsinger19103: What is (E) ?

specs/pvs/corla.pvs
```diff
@@ -981,6 +999,8 @@ BEGIN
   IMPORTING elections,
             FAFSSL
 
+%%% Audit board members are members of the public who come from different political parties. According to 
+
```
sfsinger19103: According to ... what? And where is the distinct-party criterion enforced?

dmzimmerman:
@sfsinger19103, yes, that definitely has to change. Sadly, the CVR files do not give us sufficient information to change it, so we'd actually _need_ to use the election definition files to do so.

specs/pvs/corla.pvs
dmzimmerman: So far, nowhere. It would be enforced by an invariant check on the audit board selection endpoint, but right now, political party is just a String... so if one is "REPUBLICAN" and the other is "Republican!", that'd be considered different. We may need a table of political parties to select from (for phase 2 or 3, not phase 1).

nealmcb:
@sfsinger19103 Are you aware of contests in colorado for which the number of winners differs from the Vote for information? I'm not, though I know the VVSG CVR standard handles it, e.g. for Approval Voting.

sfsinger19103:
I won one! Philadelphia City Commissioner, General Election, is
vote-for-two with three winners.

Same for most County Commissioner races in Pennsylvania.

Philadelphia City Council is vote-for-five with seven winners.

These rules protect the body from full control by one party.

Stephanie Singer
Principled Project Lead
Free & Fair <http://freeandfair.us>
267-414-3119

specs/pvs/corla.pvs
```diff
@@ -13,15 +13,31 @@
...
-%%% that document for our purposes. We will call that document
-%%% RDPR-6-Jul-2017 for the purpose of traceability in this
-%%% specification.
+%%% {Section 1-7-515, C.R.S.}, where ``C.R.S.'' denotes ``Colorado Revised Statutes''.
+%%%
+%%% The rules pertaining to election audits are found in Code of
+%%% Colorado Regulations, Secretary of State, State of Colorado,
+%%% Department of State Elections Rules
+%%% \href{http://www.sos.state.co.us/CCR/GenerateRulePdf.do?ruleVersionId=7121&fileName=8%20CCR%201505-1}{\textbf{8
+%%% CCR 1505-1}}. We will call that document \emph{Rules 1505-1} for
```
nealmcb: That version is not yet updated.
For today at least, Rule 25 is here:
http://www.sos.state.co.us/pubs/rule_making/CurrentRules/8CCR1505-1/Rule25.pdf
and others are at http://www.sos.state.co.us/pubs/rule_making/CurrentRules/8CCR1505-1Elections.html

specs/pvs/corla.pvs
kiniry: Left over from previous edits.  I have removed it.

specs/pvs/corla.pvs
kiniry: According to CDOS via email, from what I recall.  We'll have to probably canonicalize parties or add a couple of buttons to represent D/R to ensure that this isn't screwed up.  Alternatively, we just make the statement and leave it on the audit boards and the SOS to make the determination.  (I like the latter solution better.  We'll ask CDOS how they want it.  I'll file an issue with the appropriate labels.)

specs/pvs/corla.pvs
kiniry: Fixed.

PR #294, opened 2017-08-16
Two-Factor Authentication

### LINKED ISSUE #282: Implement Authentication
Right now, the implementation of authentication checks for a valid administrator account in the database, but does not check passwords or a second factor. Both of those need to be added.

kiniry:
This PR's work will contain the formal specification and façade implementation of two-factor authentication.  It also includes our test mode implementation.  Closes #282, #294.


kiniry:
I'm going to start testing this branch now.

kiniry:
Traditional authentication to the DB works now.  The implementation supporting CDOS's proprietary 2FA is also done and is being put into a private repo.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
```diff
@@ -0,0 +1,237 @@
...
+                                        final String the_username) {
+    the_request.session().removeAttribute(ADMIN);
+    Main.LOGGER.info("session is now traditionally deauthenticated");
+  }
+  
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void twoFactorDeauthenticate(final Request the_request,
```
dmzimmerman: I'm not certain what the need is for two different types of deauthentication. Are they meant to be overridden by child classes? It seems like it would be odd to let child classes mess with the session handling.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
```diff
@@ -0,0 +1,237 @@
...
+   * @param the_request The request.
+   * @param the_admin_type the type of administrator to attempt to authenticate.
+   * @param the_username the username of the person to attempt to authenticate.
+   * @param the_password the password for `username`.
+   * @param the_second_factor the second factor for `username`.
+   */
+  @Override
+  public boolean authenticateAdministrator(final Request the_request,
+                                           final Response the_response,
+                                           final AdministratorType the_admin_type,
```
dmzimmerman: We probably want to eliminate the AdministratorType from this parameter list at this point, since it looks to me as though it is in fact not used for anything.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
```diff
@@ -0,0 +1,237 @@
...
+        the_request.session().attribute(AuthenticationInterface.AUTH_STAGE);
+    return auth_stage != null && 
+        auth_stage.equals(AuthenticationInterface.ADMIN);
+  }
+  
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void logoutAdministrator(final Request the_request,
```
dmzimmerman: What is this method for (i.e., why not just use `deauthenticate`)?

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
```diff
@@ -0,0 +1,237 @@
...
+  @Override
+  public boolean isAuthenticatedAs(final Request the_request,
+                                   final AdministratorType the_type,
+                                   final String the_username) {
+    boolean result = false;
+    final Object admin_attribute = the_request.session().attribute(ADMIN);
+    if (admin_attribute instanceof Administrator) {
+      final Administrator admin = (Administrator) admin_attribute;
+      result = admin.type() == the_type;
+      the_request.session().attribute(ADMIN, admin);
```
dmzimmerman: I'm not sure why you're putting the object back in the session again here... you didn't remove it, so it should still be there, and you haven't changed it.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
dmzimmerman: I'm also not sure why the `deauthenticate` method needs a username parameter. I always thought of it as "deauthenticate the session", so it doesn't really matter who the session was authenticated as.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/Authentication.java
```diff
@@ -68,22 +64,8 @@ private Authentication() {
...
-    } else if (admin_attribute != null) {
-      // this should never happen since we control what's in the session object,
-      // but if it does, we'll clear out that attribute and thereby force another
-      // authentication
-      Main.LOGGER.error("Invalid admin type detected in session.");
-      unauthenticate(the_request);
-    }
-    
-    return result;
+    assert false;
```
dmzimmerman: Should I interpret this to mean "this class is now deprecated"? Which is fine... but why not, then, just make the appropriate changes to the parts of the system that use it and remove it? Or, alternatively, leave it alone until things are switched over?

dmzimmerman: (I see you did make appropriate changes to parts of the system that use it... so you might as well delete it - that'll tell you for sure if there are parts of the system you missed, for one thing :) )

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AuthenticateCountyAdministrator.java
```diff
@@ -94,16 +94,20 @@ protected String asmIdentity(final Request the_request) {
...
-        final SubmittedUsernamePassword auth_info = 
-            Main.GSON.fromJson(the_request.body(), SubmittedUsernamePassword.class);
-        if (auth_info != null &&
-            Authentication.authenticateAs(the_request, auth_info, 
-                                          AdministratorType.COUNTY)) {
-          ok(the_response, "Authenticated");
+        if (Main.authentication().
+            authenticateAdministrator(the_request, the_response,
+                AdministratorType.COUNTY, 
+                the_request.queryParams(AuthenticationInterface.USERNAME),
```
dmzimmerman: Why are you changing the style of submission from JSON request body to HTTP parameters? Is it to be able to present a relatively standard HTML form? I suspect @ranweiler would probably prefer a JSON request body, since that's what he had me change the old way to in the first place (though I kept the HTTP parameters around so we could use the static web page for testing).

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Administrator.java
```diff
@@ -87,16 +88,14 @@
...
-  @Column(nullable = false, updatable = false)
-  private String my_two_factor_auth_info;
+  private Instant my_last_login_time;
   
   /**
-   * The last login time.
+   * The last logout time.
    */
-  private Instant my_last_login_time;
+  private Instant my_last_logout_time;
```
dmzimmerman: Ah... is it to update the last logout time that you want a username submitted to the logout/deauthenticate endpoint? I still think it's unnecessary - you get the actual administrator object from the session, and can just set it then.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Administrator.java
```diff
@@ -191,13 +183,27 @@ public void updateLastLoginTime() {
...
+  
+  /**
    * @return a String representation of this contest.
    */
   @Override
   public String toString() {
     return "Administrator [username=" + my_username + ", type=" + 
            my_type + ", full_name=" + my_full_name + ", last_login_time = " + 
-           my_last_login_time + "]";
+           my_last_login_time + ", last_logout_time = " + my_last_logout_time + "]";
```
dmzimmerman: By convention I have not been putting spaces in the `toString` around the equals signs (though apparently `last_login_time` was an exception to that rule).

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
kiniry: The design is driven by the fact that we may someday support:
 - being authenticated to multiple roles simultaneously, and
 - some authentication systems force one to identify the credential being de/un-authenticated, though many do not

And yes, the two different kinds of authentication and de-authentication are meant to be overridden by child classes.  You'll see that in the Entrust stuff it matters.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
kiniry: I kept it because I imagine some deployments may be using different authentication server and services for different roles.  E.g., public might use an HTTP mechanism, while administrators might use some kind of 2FA.  Indeed, we are not using it for this particular deployment.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
kiniry: Indeed, it is a refactoring artifact.  I've removed it in e6c81bb.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
kiniry: That's a line straight from your implementation.  I was puzzled by it too.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/Authentication.java
kiniry: Yes, I was testing with it bottomed out.  It is now going away.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AuthenticateCountyAdministrator.java
kiniry: Can/should we support both, given the continued need for the static page for testing?

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
dmzimmerman: Well, let's get rid of it. I suspect it was a cut and paste error on my part that nobody caught in review, to be honest. :)

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Administrator.java
kiniry: Yes, that is the purpose, though I was still on the fence about its utility.  You'll see the design didn't migrate over into `AbstractAuthentication`.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AuthenticateCountyAdministrator.java
dmzimmerman: I think it's a reasonable thing for us to continue to support.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Administrator.java
kiniry: Got it; for the future.  That class is gone now, FWIW.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Administrator.java
dmzimmerman: I got rid of it, but `deauthenticate` does still pull the username from the active session if there is one when it calls the child classes.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Administrator.java
dmzimmerman: Um... I doubt it. If you deleted the `Administrator` class, we'd have serious problems. :)

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
```diff
@@ -59,78 +59,95 @@
...
-            Main.LOGGER.info("Second factor authentication succeeded for administrator " + 
-                the_username);
-          } else {
-            // Send the authentication state machine back to its initial state.
-            the_request.session().attribute(AUTH_STAGE, NO_AUTHENTICATION);
-            Main.LOGGER.info("Second factor authentication failed for administrator" + 
-                the_username);
-            result = false;
-          }
+        switch (auth_stage) {
```
kiniry: This is definitely cleaner.  Thanks.

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
```diff
@@ -243,55 +291,84 @@ public boolean isAuthenticatedAs(final Request the_request,
...
-            the_username.equals(admin.username());
-      } else if (admin_attribute != null) {
-        // this should never happen since we control what's in the session object,
-        // but if it does, we'll clear out that attribute and thereby force another
-        // authentication
-        Main.LOGGER.error("Invalid admin type detected in session.");
-        deauthenticate(the_request, the_username);
-      }
+        admin_attribute instanceof Administrator) {
+      final AuthenticationStage stage = (AuthenticationStage) auth_stage_attribute;
```
kiniry: stage can be null here, right?  I guess it won't hurt given the below comparison to the enum?

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AbstractAuthentication.java
```diff
@@ -327,17 +404,22 @@ public final SubmittedCredentials authenticationCredentials(final Request the_re
     // Check for JSON credentials in the request.
     try {
       result = Main.GSON.fromJson(the_request.body(), SubmittedCredentials.class);
```
kiniry: So this isn't where the body of the request was getting flushed? 😕

server/eclipse-project/src/main/java/us/freeandfair/corla/auth/AuthenticationInterface.java
```diff
@@ -115,13 +116,12 @@ boolean secondFactorAuthenticate(Request the_request,
...
-   * Is `the_username` authenticated with a second factor?
+   * Is the session authenticated with a second factor?
    * @trace authenticated.second_factor_authenticated?
    * @param the_request The request.
-   * @return true iff `the_username` is second-factor authenticated.
+   * @return true iff the session is second-factor authenticated
    */
-  boolean secondFactorAuthenticated(Request the_request,
-                                    String the_username);
+  boolean secondFactorAuthenticated(Request the_request);
```
kiniry: Well, looks like I lost that design battle. 😁

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -397,6 +397,10 @@ public void before(final Request the_request, final Response the_response) {
     reset();
     my_log_entries.set(new ArrayList<LogEntry>());
     Main.LOGGER.info("endpoint " + endpointName() + " hit by " + the_request.host());
+    // make sure we get all the HTTP post parameters, if there are any, before
+    // anything has a chance to read the request body before Spark
+    the_request.queryParams();
```
kiniry: That's it?!

dmzimmerman: Amazingly, yes.

PR #356, opened 2017-08-18
Help build the deliverable zip: build_release etc

nealmcb:


nealmcb:
TODO: add in INSTALL.md and README.md files for the release, which aren't actually in the most recent deliverable, so perhaps @dmz can add them. What directory should those go in?

The build_release script also needs to build html files for each of those, along with 
 runbook.pdf, user_manual.pdf, corla_model.pdf, book.pdf
and probably put some or all of them in the apache root under docs or something.

nealmcb:
TODO: add apache config files, including changes to redirect `/login` etc the client, ala `ErrorDocument 404 /index.html`

nealmcb:
Given the corla.conf file from dmz which I've added, this is now suitable for review and merging to make deployment easier, and also make testing and merging a more functional docker-support easier.
Building the .pdf files would be a worthy addition, but lower priority.

PR #370, opened 2017-08-20
Improved Performance

dmzimmerman:
This set of code improves performance of the server dramatically (as Neal mentioned in the performance ticket), but does not introduce the rearchitected audit functionality (I'm still ironing some issues out of that, which is not helped by not being able to smoketest locally at home).


server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
```diff
@@ -89,7 +89,7 @@
   /**
    * The county ID to apply to the parsed manifest lines.
    */
-  private final Integer my_county_id;
+  private final Long my_county_id;
```
kiniry: What prompted this change in primitive type?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/SelectContestsForAudit.java
```diff
@@ -71,18 +70,18 @@ public synchronized String endpoint(final Request the_request,
...
-      final DoSDashboard dosdb = DoSDashboardQueries.get();
+      final DoSDashboard dosdb = Persistence.getByID(DoSDashboard.ID, DoSDashboard.class);
       if (dosdb == null) {
         Main.LOGGER.error("could not get department of state dashboard");
         serverError(the_response, "Could not select contests");
       } else {
         for (final ContestToAudit c : contests) {
           Main.LOGGER.info("updating contest audit status: " + c);
           dosdb.updateContestToAudit(c);
+          Persistence.saveOrUpdate(dosdb);
```
kiniry: Was this a serious bug?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CVRAuditInfo.java
```diff
@@ -0,0 +1,151 @@
...
+   * 
+   * @param the_other The other object.
+   * @return true if the objects are equivalent, false otherwise.
+   */
+  @Override
+  public boolean equals(final Object the_other) {
+    boolean result = true;
+    if (id() != null && the_other instanceof CVRAuditInfo) {
+      final CVRAuditInfo other_info = (CVRAuditInfo) the_other;
+      // this is a shortcut; since there's a unique constraint on
```
kiniry: A tricky optimization to track if/when the schema semantics change.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CVRContestInfo.java
```diff
@@ -87,7 +87,7 @@
   // this is a list of choice names to make persistence more straightforward; if it
   // were a list of Choice, then the mapping between contests and choices would
   // need to be more complex
-  @ElementCollection(fetch = FetchType.EAGER)
+  @ElementCollection(fetch = FetchType.LAZY)
```
kiniry: Do you yet have a rule-of-thumb for EAGER vs LAZY in Hibernate?

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
dmzimmerman: Uniformity in database table "ID" columns being represented as java.lang.Long. At this point I probably _could_ switch the county and dashboard ones to Integer (since they are no longer inheriting from the same parent), but I don't know that I see the point.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/SelectContestsForAudit.java
dmzimmerman: No, I don't think so. And in fact, a future PR is probably going to rip out a lot of those persistence calls in favor of letting Hibernate just do its job.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CVRAuditInfo.java
dmzimmerman: Actually, these entities are entirely owned by their creating entities, and comparing them for equivalence by ID is unambiguously the right thing to do with them. The comment there is misleading because actually, that unique constraint doesn't exist anymore. I could rip out that comment before merging this PR, I guess.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CVRContestInfo.java
dmzimmerman: Not really. LAZY wherever LAZY works, except for very small collections where the performance benefit of loading them into memory all at once (using one query) outweighs the overhead of that memory usage. 

I'm looking to rework part of the `Persistence` class to make session handling and transaction handling more controlled by us and less by depending on Hibernate's `getCurrentSession()` - I've seen some weirdness where sessions end before they should in the current scheme (which makes LAZY not work in all cases). That will simplify a _lot_ of the transaction setup and such, because it will make `AbstractEndpoint` the sole class that ever begins or ends a transaction, and we'll just say on the various query classes and everything that you need to be in a transaction to invoke them. I also need to examine some of the places where I'm using IDs instead of regular object references (such as `CVRAuditInfo`) to see if lazy loaded object references would make more sense, and do a bit of refactoring in the model classes; there's too much use of `Persistence` creeping into the model classes again. But I digress...

Issue #364, opened 2017-08-18
submitting a bad hash for an upload should result in a clear error message


kiniry:
In particular, the County Administrator should be given guidance on how to double-check and remedy the situation.  Right now pushing "Upload" using a bad (or missing) hash is a UI NOP.

ranweiler:
We now have popups indicating an error, but need to make the errors more specific/include guidance.

PR #376, opened 2017-08-21
File upload UX

ranweiler:
Towards #364. We need to add more info, but this is a start.


ranweiler:
@kiniry yes, that's correct. Actually, I think I can improve that with a small change, hold please.

ranweiler:
@kiniry okay, I was incorrect: differentiating with a nice display of the reasons for failure is enough work that I'd rather get this in, and improve it in another PR tomorrow.

kiniry:
I'm ok with that, so long as an issue is/was filed.

ranweiler:
@kiniry yeah, I'm just leaving #364 open.

PR #388, opened 2017-08-22
Add 5 Denver CVRs

nealmcb:


nealmcb:
We already have permission from Dwight to include the csv version of these cvrs. We're waiting on permission for the JSON-based data.

PR #391, opened 2017-08-22
Transaction Simplification

dmzimmerman:
This PR dramatically simplifies the way JPA/Hibernate sessions and transactions are handled, including making uploads transactional (though still not using the model we would prefer to be using, which should happen later but requires client changes). It includes proper usage of the Hibernate second-level cache, proper connection pooling, and a sweeping removal of transaction-related calls everywhere except `AbstractEndpoint`.


server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
```diff
@@ -126,6 +137,20 @@ public ColoradoBallotManifestParser(final String the_string,
   }
   
   /**
+   * Checks to see if the set of parsed manifests needs flushing, and does so 
+   * if necessary.
+   */
+  private void checkForFlush() {
+    if (my_parsed_manifests.size() % BATCH_SIZE == 0) {
```
ranweiler: Would it be better to check `> BATCH_SIZE`? As written, this will not flush for any multiple of `BATCH_SIZE`.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
```diff
@@ -207,20 +228,14 @@ public synchronized boolean parse() {
     my_parse_status = true;
     my_parse_success = result;
     
-    // if we had any kind of parse error, we scrap the whole import
-    
-    if (!result) {
```
ranweiler: What is now reponsible for the equivalent of `Persistence.delete($Class, $id)`, or how is it sidestepped? I see that below you were able to remove the other `abort()` methods, too.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
```diff
@@ -223,6 +235,20 @@ private void addContests(final CSVRecord the_choice_line,
   }
   
   /**
+   * Checks to see if the set of parsed CVRs needs flushing, and does so 
+   * if necessary.
+   */
+  private void checkForFlush() {
+    if (my_parsed_cvrs.size() % BATCH_SIZE == 0) {
```
ranweiler: See comment above about how this is being checked.

More generally, can `BATCH_SIZE` and some of this parsing logic be factored out?

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
```diff
@@ -361,17 +382,17 @@ public synchronized boolean parse() {
...
       my_county.contests().addAll(my_contests);
       Persistence.saveOrUpdate(my_county);
-    } else {
-      abort();
     }
+    
     return result;
   }
-
+
```
ranweiler: Added new stray whitespace. Not worth a commit, but we should really trim trailing whitespace on all the docs and Java files.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -344,12 +341,19 @@ public void before(final Request the_request, final Response the_response) {
...
+   * The afterAfter filter for this endpoint. By default, it attempts to commit 
    * any open transaction (this makes writing endpoint code more straightforward,
    * as the vast majority of endpoints will never have to deal with transactions
    * themselves).
    */
-  public void after(final Request the_request, final Response the_response) {
+  public void afterAfter(final Request the_request, final Response the_response) {
     // try to take the transition for this endpoint in the ASM and save it to the DB
     // note that we do not try to commit when we have an error code in the response
     if (the_response.status() == HttpStatus.OK_200 &&
```
ranweiler: Not necessarily for this PR, but this should probably check that the status is 2xx, not 200. For example, a 201 would be appropriate for the success path of many of our endpoints.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -344,12 +341,19 @@ public void before(final Request the_request, final Response the_response) {
...
+  }
+  
+  /**
+   * The afterAfter filter for this endpoint. By default, it attempts to commit 
    * any open transaction (this makes writing endpoint code more straightforward,
    * as the vast majority of endpoints will never have to deal with transactions
    * themselves).
    */
-  public void after(final Request the_request, final Response the_response) {
+  public void afterAfter(final Request the_request, final Response the_response) {
```
ranweiler: Not a fan of this name. Can we call it anything more specific to what it is meant for?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
```diff
@@ -545,6 +557,42 @@ public Integer estimatedBallotsToAudit() {
...
+  }
+
+  /**
+   * Compare this object with another for equivalence.
+   * 
+   * @param the_other The other object.
+   * @return true if the objects are equivalent, false otherwise.
+   */
+  @Override
+  public boolean equals(final Object the_other) {
```
ranweiler: Idea: couldn't we push the dynamic type checking of `Object` into a public `equals` method on a new `AbstractModel` base class, and then have that call a method provided by the implementing class which only ever compares two of the given model? A similar strategy might help make a lot of the model code easier to change and review.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/RLAAlgorithm.java
```diff
@@ -147,19 +148,22 @@ public RLAAlgorithm(final CountyDashboard the_dashboard) {
...
                                             my_dashboard.id(),
-                                            RecordType.UPLOADED) - 1;
+                                            RecordType.UPLOADED);
+    if (!count.isPresent()) {
+      throw new IllegalStateException("unable to count CVRs for county " + my_dashboard.id());
+    }
 
-    final int maximum = max_long.intValue();
+    final boolean with_replacement = true;
+    // assuming that CVRs are indexed from 0
```
ranweiler: :+1: this is a better place to do this.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/RLAAlgorithm.java
```diff
@@ -147,19 +148,22 @@ public RLAAlgorithm(final CountyDashboard the_dashboard) {
    * @param the_cvr_count the total number of CVRs in a given county.
    */
   public List<Long> computeBallotOrder(final String the_seed) {
-    final boolean with_replacement = true;
-    // assuming that CVRs are indexed from 0
-    final int minimum = 0;
     // the number of CVRs for the_contest_to_audit - note that the sequence
```
ranweiler: Maybe move this note down below to the new place where you subtract 1.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
dmzimmerman: It doesn't need to exist anymore because... transactional.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
dmzimmerman: It would not. It doesn't matter whether it flushes for any multiple of batch size, because we don't care - the transaction will be committed after that. This is just to make sure we don't keep too much in memory at once.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
dmzimmerman: I don't think it's worth doing at the moment, since it occurs only in two unrelated parsers that don't really need to have the same batch size, but maybe in the future...

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
dmzimmerman: No. It is an implementation of a Spark `afterAfter` hook.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
dmzimmerman: We never send anything other than a 200, but I can certainly modify the check for this.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
dmzimmerman: That would be very un-Java-conventional. And since Java doesn't support multiple inheritance, probably pretty messy, since the models don't currently all share the same base class.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/RLAAlgorithm.java
dmzimmerman: Good ides.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
dmzimmerman: I've already refactored this in a later PR so I'm going to leave it as is here.

ranweiler: :+1:

PR #407, opened 2017-08-23
ASM Revisions

dmzimmerman:
We want counties to be able to upload things and select an audit board in any order. This update to the state machines accomplishes that.


kiniry:
I'm scared the @dmzimmerman can layout state machines in his head.

PR #413, opened 2017-08-23
New main.py with command-line arguments.

nealmcb:
Worth merging I think, whenever you want it.

~Currently requires a county option, like "./main.py -c 3" for Arapahoe.~

~Not quite working yet: never sees any ballots to audit for some reason.~

### DIFF
--- a/test/smoketest/bootstrap_cvr.py
+++ b/test/smoketest/bootstrap_cvr.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+"""Bootstrap a Dominion CVR file
+Adapted from 
+ Bootstrap a very large data file | Tomochika Fujisawa's site
+ https://tmfujis.wordpress.com/2016/06/24/bootstrap-a-very-large-data-file/
+"""
+
+import sys
+import numpy
+
+for line in open(sys.argv[1], "r"):
+        cnt = numpy.random.poisson(lam=1)
+
+        if cnt == 0:
+                continue
+        else:
+                for i in range(cnt):
+                        sys.stdout.write(line)
--- a/test/smoketest/main.py
+++ b/test/smoketest/main.py
@@ -92,9 +92,6 @@ __license__ = "TODO GPL"
 
 parser = argparse.ArgumentParser(description='Drive testing for ColoradoRLA auditing.')
 
-parser.add_argument('-R, --reset', dest='reset', action='store_true',
-                    default=False,
-                    help='reset the database before other actions')
 parser.add_argument('-c, --county', dest='counties', metavar='COUNTY', action='append',
                     type=int,
                     help='numeric county_id for the given command, e.g. 1 '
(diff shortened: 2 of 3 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/AbstractEndpoint.java
```diff
@@ -55,7 +55,7 @@
   /**
    * A flag that disables ASM checks, when true.
    */
-  public static final boolean DISABLE_ASM = false;
+  public static final boolean DISABLE_ASM = true;   // FIXME: should be false, Do not let "true" in to master...
```
dmzimmerman: Let's fix this.

test/smoketest/bootstrap_cvr.py
```diff
@@ -0,0 +1,18 @@
...
+import numpy
+
+for line in open(sys.argv[1], "r"):
+        cnt = numpy.random.poisson(lam=1)
+
+        if cnt == 0:
+                continue
+        else:
+                for i in range(cnt):
+                        sys.stdout.write(line)
```
dmzimmerman: Is this file used elsewhere for something? It looks like the start of a CVR file generator.

test/smoketest/main.py
```diff
@@ -186,35 +296,65 @@ def publish_ballots_to_audit(seed, cvrs):
...
-                    "413befb5bc3e577e637bd789a92da425d0309310c51cfe796e57c01a1987f4bf")
-    if v == 2:
-        upload_cvrs(baseurl, s, "../dominion-2017-CVR_Export_20170310104116.csv",
-                    "4e3844b0dabfcea64a499d65bc1cdc00d139cd5cdcaf502f20dd2beaa3d518d2")
+    manifest = ac.args.manifest # TODO finish when we have more manifests and they make a difference
 
-    if v == 3:
-        upload_cvrs(baseurl, s, "../Denver2016Test/CVR_Export_20170804111144.csv",
-                    "1def4aa4c0e1421b4e5adcd4cc18a8d275f709bc07820a37e76e11a038195d02")
+    if ac.args.ye_olde_upload:
```
dmzimmerman: I would remove this argument and logic. The old upload endpoints will go away soon, and there isn't much point in doing further testing on them; moreover, if somebody for some reason wants to test an old version, they can use the smoketest script that's current at that version in the repo.

test/smoketest/bootstrap_cvr.py
nealmcb: Yes - I used it (with some other head and tail commands) to make what we're testing with. Next on my plate is fixing the CVR ids as you requested.

test/smoketest/main.py
nealmcb: For now I still want to be able to test old branches, with the flexibility of this code, e.g. for timing, and hate to be without the code that is necessary.  I'll remove them later. And I like the name, go figure.

test/smoketest/bootstrap_cvr.py
dmzimmerman: OK, cool.

PR #420, opened 2017-08-24
Serialize Instants in ISO-8601 Format

### LINKED ISSUE #382: Change JSON Serialization Format for Instant
Instead of seconds and nanos, we should send Instants as a human readable string. I suggest something like the ISO 8601 Instant format, which would render them as strings like `2007-12-03T10:15:30.00Z` (and has the additional benefit of being parsable by the Instant class for reconstituting them).

dmzimmerman:
Closes #382


nealmcb:
Excellent! LGTM, and seems to work fine.

Issue #467, opened 2017-08-25
Missed Deadline and Audit In Progress


dmzimmerman:
When a county ASM is put into the `MISSED_DEADLINE` state, the corresponding audit board ASM should not be put into the `AUDIT_IN_PROGRESS` state; it should be left in `AUDIT_INITIAL_STATE`.

dmzimmerman:
Actually, I think the audit board state machine needs an `UNABLE_TO_AUDIT` state (or similar), that is a final state that is reached when the county misses its deadline.

PR #468, opened 2017-08-25
Fix Server State Transitions at Audit Start

dmzimmerman:
Previously, if a county missed the upload or audit board establishment deadline, their audit board ASM would be transitioned into the `AUDIT_IN_PROGRESS` state anyway as soon as ballots to audit were published. Now, they are only transitioned into that state if the county ASM is transitioned successfully into the `COUNTY_AUDIT_UNDERWAY` state.


dmzimmerman:
Addresses the server side of #471
Closes #470 
Closes #467

server/eclipse-project/src/main/java/us/freeandfair/corla/json/DoSDashboardRefreshResponse.java
```diff
@@ -168,7 +168,7 @@ private static Integer ballotsToAudit(final Contest the_contest) {
                    Math.max(0, ccca.ballotsToAudit() - 
                                ccca.dashboard().auditedPrefixLength()));
     }
-    return to_audit;
+    return Math.max(0, to_audit);
```
kiniry: I like that fix.

PR #477, opened 2017-08-25
Change CVR Storage Format and Upload Mechanism for Efficiency and Better Typing

### LINKED ISSUE #228: Convert TabulatorNum, BatchId, and RecordId from CVR: string to integer
The fields `TabulatorNum`,`BatchId`,`RecordId`,`ImprintedId` in a Dominion 5.2 CVR in csv format are all represented as strings. This interpretation is reinforced by the `ImprintedId` field, shown for example for a CVR with scanner id 3, batch 800 and sequence number 3 as `="3-800-3"` in the sample data provided. Note that the spurious `=` used in these fields is a separate topic being reviewed by Dominion.

See [`test/dominion-2017-CVR_Export_20170310104116.csv`](https://github.com/FreeAndFair/ColoradoRLA/blob/1ba906a8c5c69ff2d23c0b6342259756638ceb83/test/dominion-2017-CVR_Export_20170310104116.csv) for an example.

They are thus stored as strings in the RLA tool database.  

But representing them as strings makes little sense, especially for `RecordId`, which is a numerical sequence number of a ballot in a batch.  This is an issue especially when sorting them into order, both for verifiable random selection and for presenting the randomly selected ballots to the counties for retrieval. Counties probably want to see batch "10" after batch "9", not after batch "1".

The types of these fields are not specified in the _CVR export file format_ documentation from Dominion. But a closer examination of the JSON version of these files shows that these fields are represented as integers there. (Note that `TabulatorNum` is shown as `TabulatorId` in the JSON files, and there is no `ImprintedId` field there.)

Given that these are ints in the more-canonical JSON formats, these fields in the CVR.csv file should probably be converted to int when read in, and the RLA tool should reject any CVR or manifest which has non-numeric data in these fields. Then the records can easily be properly sorted.

Until this change is made, it may be possible to generate CVRs and manifests in which these ids show up with leading zeros, so that sorting by string will produce a natural numeric order.

This proposed resolution is pending review by the Department of State and Dominion.

dmzimmerman:
Having the CVR ID stored in the database gives us a natural ordering on CVRs for random selection, which should reduce the amount of work the database needs to do and probably the amount of information we need to store/load at a time. 

Among other things, closes #228


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
```diff
@@ -0,0 +1,170 @@
...
+   * {@inheritDoc}
+   */
+  @Override
+  @SuppressWarnings({"PMD.ConfusingTernary"})
+  public String endpoint(final Request the_request, final Response the_response) {    
+    // we know we have county authorization, so let's find out which county
+    final County county = Authentication.authenticatedCounty(the_request);
+
+    if (county == null) {
+      unauthorized(the_response, "unauthorized administrator for CVR epxort upload");
```
ranweiler: s/epxort/export

server/eclipse-project/src/main/java/us/freeandfair/corla/json/UploadedFileJsonAdapter.java
```diff
@@ -0,0 +1,138 @@
...
+import com.google.gson.stream.JsonWriter;
+
+import us.freeandfair.corla.model.UploadedFile;
+import us.freeandfair.corla.persistence.Persistence;
+
+/**
+ * JSON adapter for uploaded files. This enapsulates in JSON all the information
+ * about the file except for its actual contents, and retrieves from the database
+ * the uploaded file corresponding to the specified county and database ID 
+ * ignoring other fields; note that "null" can be retrieved if there is no
```
ranweiler: s/"can be retrieved"/"can be returned"

server/eclipse-project/src/main/java/us/freeandfair/corla/json/UploadedFileJsonAdapter.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+  
+  /**
+   * Reads an uploaded file object.
+   * 
+   * @param the_reader The JSON reader.
+   * @return the object.
+   */
+  @Override
+  public UploadedFile read(final JsonReader the_reader)
```
ranweiler: What's going on with this method? Did you mean to start implementing this?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
```diff
@@ -0,0 +1,170 @@
...
+      try {
+        Persistence.saveOrUpdate(cdb);
+      } catch (final PersistenceException e) {
+        serverError(the_response, "could not update county dashboard");
+      }
+    }
+  }
+  
+  /**
+   * Parses an uploaded ballot manifest and attempts to persist it to the database.
```
ranweiler: s/ballot manifest/CVR export

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
```diff
@@ -0,0 +1,170 @@
...
+   * {@inheritDoc}
+   */
+  @Override
+  protected ASMEvent endpointEvent() {
+    return UPLOAD_CVRS_EVENT;
+  }
+
+  /**
+   * Updates the appropriate county dashboard to reflect a new 
+   * ballot manifest upload.
```
ranweiler: s/ballot manifest/CVR export

server/eclipse-project/src/main/java/us/freeandfair/corla/json/UploadedFileJsonAdapter.java
```diff
@@ -0,0 +1,142 @@
...
+  public void write(final JsonWriter the_writer, 
+                    final UploadedFile the_file) 
+      throws IOException {
+    the_writer.beginObject();
+    the_writer.name(FILE_ID).value(the_file.id());
+    the_writer.name(COUNTY_ID).value(the_file.countyID());
+    the_writer.name(FILENAME).value(the_file.filename());
+    the_writer.name(SIZE).value(the_file.size());
+    the_writer.name(TIMESTAMP).value(the_file.timestamp().toString());
+    the_writer.name(HASH_STATUS).value(the_file.hashStatus().toString());
```
ranweiler: We should be including the actual computed hash here too, right?

dmzimmerman: No. That would allow them to just use that to resubmit the same file.

server/eclipse-project/src/main/java/us/freeandfair/corla/json/UploadedFileJsonAdapter.java
dmzimmerman: I find myself wondering if I failed to push a commit to this branch.

dmzimmerman: Um... no, the whole method is there. What's going on with your comment? :)

ranweiler: I thought maybe you hadn't finished implementing this. For example:
- Switch cases without breaks and what seem to be fallthroughs that we don't want. Are you just batching okay-to-ignore keys with a fallthrough case?
- Specifically, I was confused by skipping the filename field if we have one.

kiniry: Looks fine to me.

dmzimmerman: If you're wondering why everything other than file ID is ignored, the answer is, because it's irrelevant to getting the file with that ID out of the database and we're checking that the correct county owns that file before importing it anyway.

dmzimmerman: (i.e., it's to make the client's life potentially easier by having to submit a smaller file descriptor object, namely `{ "id": blah }`), but also being able to just submit the reply verbatim.

ranweiler: @dmzimmerman that explanation makes sense.

Issue #331, opened 2017-08-17
ensure glossary is up-to-date, appropriate, and consistent for User's Guide, Tutorial, and Specification


kiniry:

cjerdonek:
A couple suggestions:

* add a few terms around open source (e.g. "open source," "copyleft," etc)
* alphabetize the entries

cjerdonek:
PS - since the glossary is big, it looks like it would be easy to script alphabetizing it (e.g. in Python by splitting on `\n* **`, sorting, and then joining again).

kiniry:
Thanks for your input @cjerdonek.  @mkiniry, make sure to add those terms and work with us to get the full glossary done for the `phase-3` release.  

And yes, @cjerdonek, we did intend to auto-sort the list.  Feel free to provide a script in a PR. ;)

cjerdonek:
I added a script in PR #481. You don't need to check it in if you just want to use it once.

sfsinger19103:
There are several @review comments still in the glossary. What procedure do we want to use to resolve them? E.g., @sfsinger resolves them using her best judgment and asks for review from @kiniry?

PR #481, opened 2017-08-26
Add script to sort glossary entries.

cjerdonek:
As discussed [here](https://github.com/FreeAndFair/ColoradoRLA/issues/331#issuecomment-324986921).

I don't know where you wanted the script, so I just put it at the top level. Also, I noticed there's a redundant file `docs/glossary_concepts.txt`, so I removed it.


sort-glossary.py
```diff
@@ -0,0 +1,32 @@
...
+        text = f.read()
+
+    sep = '\n\n* **'
+    parts = text.split(sep)
+    # Normalize the amount of space between entries.
+    parts = [part.strip() for part in parts]
+
+    # Keep the intro at the beginning.
+    first_part = parts[0]
+    parts = sorted(parts[1:])
```
kiniry: I think that this should probably read `sorted(parts[1:], key=lambda s: s.lower())` and declare the file as Python2.

kiniry (changes requested):
Let me know what you think.  Thanks @cjerdonek!

cjerdonek:
Good point re: case-sensitivity. I half-noticed that but wasn't sure your preference. Done!

kiniry (approved):
Looks great.  Thanks, @cjerdonek!

cjerdonek:
No prob! (Btw, FWIW you can also run the script with Python 3, which is what I used.)

Issue #487, opened 2017-08-26
Ensure text fields are nonempty after trimming whitespace


ranweiler:
Disallow a validation edge case where a form text field is considered nonempty even though it contains only whitespace. Instead, trim the given values first, then check if empty.

Issue #479, opened 2017-08-26
Risk limit entry should disallow negative numbers


sfsinger19103:
Per CDOS email 8/25

PR #478, opened 2017-08-25
DB Performance

dmzimmerman:
This branch contains extreme database performance improvements, achieved through various methods including the use of database indices, reconfiguration of the entity/relationships to reduce the number of tables and joins, tuning of Hibernate settings, refinement of queries, and creative usage of type adapters to serialize lists and sets of strings in single database columns. 

Also...
Closes #365

Issue #464, opened 2017-08-25
disable SOS audit definition workflow after the audit has started


kiniry:
State Administrators should not be able to click on (or perhaps even see) the Audit menu once an audit has started.

ranweiler:
See also: #306.

ranweiler:
Closed in #490.

PR #490, opened 2017-08-26
UI fixes

ranweiler:
Includes: #478.

Closes: #487, #479, #385, #374, #306, #464.

### DIFF
--- a/client/src/component/sos/audit/AuditPage.tsx
+++ b/client/src/component/sos/audit/AuditPage.tsx
@@ -22,13 +22,18 @@ const Breadcrumb = () => (
     </ul>
 );
 
+function round(val: number, digits: number) {
+    const factor = Math.pow(10, digits);
+    return Math.round(val * factor) / factor;
+}
+
 const ReadonlyRiskLimit = ({ riskLimit }: any) => {
-    const riskLimitPercent = Math.round(riskLimit * 100);
+    const riskLimitPercent = round(riskLimit * 100, 2);
 
     return (
         <div>
             <h4>Risk limit set.</h4>
-            <div>The risk limit is set at: { riskLimit } ({ riskLimitPercent }%)</div>
+            <div>The risk limit is set at: { riskLimitPercent }%</div>
         </div>
     );
 };
--- a/client/src/component/sos/audit/AuditPage.tsx
+++ b/client/src/component/sos/audit/AuditPage.tsx
@@ -53,6 +53,8 @@ const AuditPage = ({ nextPage, riskLimit, setRiskLimit }: any) => {
                         ? <ReadonlyRiskLimit riskLimit={ riskLimit } />
                         : <RiskLimitForm forms={ forms } riskLimit={ riskLimit } />;
 
+    const buttonText = riskLimit ? 'Next' : 'Save & Next';
+
     return (
         <div>
             <Nav />
@@ -76,7 +78,7 @@ const AuditPage = ({ nextPage, riskLimit, setRiskLimit }: any) => {
             </div>
 
             <button onClick={ buttonClick } className='pt-button pt-intent-primary'>
-                Save
+                { buttonText }
             </button>
         </div>
     );
--- a/client/src/component/sos/audit/AuditReviewContainer.tsx
+++ b/client/src/component/sos/audit/AuditReviewContainer.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
+import { Redirect } from 'react-router-dom';
 
 import AuditReviewPage from './AuditReviewPage';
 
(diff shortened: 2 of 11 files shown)


client/src/component/sos/audit/AuditPage.tsx
```diff
@@ -22,13 +22,18 @@ const Breadcrumb = () => (
...
 );
 
+function round(val: number, digits: number) {
+    const factor = Math.pow(10, digits);
+    return Math.round(val * factor) / factor;
+}
+
 const ReadonlyRiskLimit = ({ riskLimit }: any) => {
-    const riskLimitPercent = Math.round(riskLimit * 100);
+    const riskLimitPercent = round(riskLimit * 100, 2);
```
kiniry: Can you explain to me the decision to use a `digits` of 2?

client/src/component/sos/audit/AuditReviewContainer.tsx
```diff
@@ -10,6 +11,10 @@ class AuditBallotListContainer extends React.Component<any, any> {
     public render() {
         const { history, sos } = this.props;
 
+        if (sos.asm.currentState === 'DOS_AUDIT_ONGOING') {
```
kiniry: Why is this conditional not in one place?

client/src/component/sos/audit/RiskLimitForm.tsx
```diff
@@ -8,63 +10,117 @@ interface FormProps {
...
-                        Ballot Polling Audits
+                        Ballot Polling Audits (%)
                         <NumericInput
-                            minorStepSize={ 0.001 }
-                            stepSize={ 0.01 }
-                            value={ ballotPollingLimit }
+                            allowNumericCharactersOnly={ true }
+                            min={ toPercent(MIN_RISK_LIMIT) }
+                            max={ toPercent(MAX_RISK_LIMIT) }
+                            minorStepSize={ toPercent(0.001) }
```
kiniry: What are the visual implications of these step sizes?

client/src/component/sos/audit/RiskLimitForm.tsx
```diff
@@ -8,63 +10,117 @@ interface FormProps {
...
                 </div>
                 <div>
                     <label>
-                        Comparison Audits
+                        Comparison Audits (%)
                         <NumericInput
-                            minorStepSize={ 0.001 }
-                            stepSize={ 0.01 }
-                            value={ comparisonLimit }
+                            allowNumericCharactersOnly={ true }
```
kiniry: Does this constraint permit one to type a decimal point?

ranweiler: @kiniry yes it does.

client/src/component/sos/audit/RiskLimitForm.tsx
ranweiler: @kiniry these determine what happens when you use click up and down arrows (`stepSize`) or shift-click (`minorStepSize`).

client/src/component/sos/audit/AuditReviewContainer.tsx
ranweiler: This is because of how we are doing URL routing. I have an idea of a better place to put it, and indeed, to put other such ASM/state based redirects, but for the moment, this is simplest approach that does what we want.

client/src/component/sos/audit/AuditPage.tsx
ranweiler: Yes. We use `digits` to compute a shift factor. The shift factor is used to shift the decimal point left or right, via division or multiplication. We truncate all sub-decimal values with `Math.round`, then shift back. So, for `d == 2`, we `factor == 100`, and for, say, `riskLimit == 1.234`, we obtain: `Math.round(1.234 * 100) / 100` -> `Math.round(123.4) / 100` -> `123 / 100` -> `1.23`.

ranweiler: Open to revision. This allows two digits past the decimal point, when regarding the risk limit as a percentage. So this allows `12.34 %`, or `0.1234`. We could narrow it if desired. I just chose the maximum precision that seemed remotely reasonable, expecting to discuss it in review.

kiniry: That's what I thought.  I certainly think that two digit precision is sufficient as a default.

client/src/component/sos/audit/AuditReviewContainer.tsx
kiniry: Ok.  Let's file an issue on that refactoring as a `someday` then.

client/src/component/sos/audit/RiskLimitForm.tsx
kiniry: Ah!  *shift*-click!  But with a 2 digit precision, one will not witness any visual feedback from such a UI action, right?

client/src/component/sos/audit/RiskLimitForm.tsx
kiniry: 👍

client/src/component/sos/audit/AuditReviewContainer.tsx
ranweiler: @kiniry filed #496.

client/src/component/sos/audit/RiskLimitForm.tsx
ranweiler: Correction: minor is for `alt`-click. We _do_ observe the expected step sizes (1% and 0.1% for `step` and `minorStep`), because I convert the steps to percentage values.

PR #495, opened 2017-08-26
Store decimals at higher precision.

### LINKED ISSUE #491: Backend should serialize risk limit with more floating point percision
The backend is either (1) accepting a risk limit with many digits of precision (e.g. `0.04321`/`4.321 %`) but not saving it with that much precision (truncates to `0.04`), or (2) is only failing to serialize the risk limit with a higher level of precision. We should be able to accept and persist risk limits with a precision of at least one tenth of a percent. In #374, @kiniry suggests that we uses 3 decimal places. I interpret this to mean we should be able to losslessly represent, for example, `12.3 %`/`0.123`.

dmzimmerman:
Closes #491

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/model/DoSDashboard.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/model/DoSDashboard.java
@@ -96,6 +96,7 @@ public class DoSDashboard implements PersistentEntity, Serializable {
   /**
    * The risk limit for comparison audits.
    */
+  @Column(precision = PRECISION, scale = SCALE)
   private BigDecimal my_risk_limit_for_comparison_audits;
   
   /**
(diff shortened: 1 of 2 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/model/DoSDashboard.java
```diff
@@ -59,6 +59,16 @@
   public static final int MIN_SEED_LENGTH = 20;
   
   /**
+   * The database stored precision for decimal types.
+   */
+  public static final int PRECISION = 6;
```
kiniry: Why the declaration here too?

ranweiler: Is it something automatically consumed by Hibernate?

dmzimmerman: It's a symbolic constant to allow us to change it more easily rather than hunting through the fields; also useful when we eventually add a risk limit for ballot polling audits, or different risk limits for state/county, etc.

dmzimmerman: Of course, it wasn't actually used yet. But it is now.

PR #497, opened 2017-08-27
Ongoing test updates: handle rounds, set-election-info, add psql tests and CVR data sets, clarify output, add options.

nealmcb:
Get bootstrap_cvr.py to automate nearly everything for creating big CVR files.

See comments in the file.


test/smoketest/bootstrap_cvr.py
```diff
@@ -1,5 +1,28 @@
...
+The CvrNumber and RecordId fields both count from 1 to the overall target number.
+An ImprintedId to match is also generated.
+
+Usage:
+
+ python bootstrap_cvr.py 100
+
+generates a 100-CVR file based on the hardcoded D0FILE constant.
+
+Note: if you don't get the desired length, you'll get a message saying to
```
kiniry: How does that happen?

nealmcb: I'm doing a very quick-and-dirty bootstrap re-sample, repeating each line a random number of times. I ask for a bit more than I expect to start with, which means I usually get at least as many as I need, but the last few lines are typically underrepresented. If the luck of the draw doesn't get to the desired length, I just ask the user to try again.
I'm sure there are lots of far more rigorous ways to get the sample I want, and given the size that I'm starting with (which all fits in memory) it wouldn't be hard. But I found this code lying around and it was cute so I went with it.....

PR #509, opened 2017-08-28
Rounds on the Server Side

dmzimmerman:
This branch implements the concept of rounds on the server side. It also modifies the `main.py` test script to properly sign in and out of audit boards at reasonable intervals, and not to do both an intermediate audit report and a final one at the end; putting an intermediate audit report in the middle somewhere might make sense instead of explicitly signing out, as intermediate audit report signs out the current audit board too.

### DIFF
--- a/test/smoketest/main.py
+++ b/test/smoketest/main.py
@@ -420,14 +420,6 @@ def county_setup(ac, county_id):
     county_s = requests.Session()
     county_login(ac, county_s, county_id)
 
-    r = test_endpoint_json(ac, county_s, "/audit-board",
-                           [{"first_name": "Mary",
-                             "last_name": "Doe",
-                             "political_party": "Democrat"},
-                            {"first_name": "John",
-                             "last_name": "Doe",
-                             "political_party": "Republican"}])
-
     upload_files(ac, county_s)
     # Replace that with this later - or make test_endpoint_file method?
     # r = test_endpoint_post(base, county_s1, "/upload-ballot-manifest", ...)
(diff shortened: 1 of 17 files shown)


test/smoketest/main.py
```diff
@@ -516,6 +516,16 @@ def county_audit(ac, county_id):
     for i in range(len(selected)):
         if i % 10 == 0:
             r = test_endpoint_get(ac, ac.state_s, "/dos-dashboard", show=False)
+            
+        if i % 40 == 0:
```
nealmcb: This should not happen as the first step. Perhaps after 5, to make it take effect in the default smoketest.

if i % 40 == 5:

test/smoketest/main.py
```diff
@@ -560,7 +570,7 @@ def county_audit(ac, county_id):
 
     r = test_endpoint_get(ac, ac.state_s, "/dos-dashboard")
 
-    r = test_endpoint_json(ac, county_s, "/intermediate-audit-report", {})
+#    r = test_endpoint_json(ac, county_s, "/intermediate-audit-report", {})
```
nealmcb: After you commit this, I'll switch the intermediate reporting to a periodic thing, and/or, when my flexible-periodic-config code lands, make it part of that.

test/smoketest/main.py
dmzimmerman: Fair enough. That's an easy change. In reality it doesn't actually matter when it occurs, I'm just checking to see that the logic works.

test/smoketest/smoketest.bash
```diff
@@ -34,7 +34,7 @@ pkill -f java.-jar.target/colorado_rla || true
 dropdb corla || true
 createdb -O corla corla
 
-mkdir -p tarket
+mkdir -p target
```
nealmcb: Thanks. I could have sworn I fixed that a while ago!
See also a fix for tabulations at https://github.com/FreeAndFair/ColoradoRLA/pull/510

PR #535, opened 2017-08-29
UI feedback

ranweiler:


client/src/component/sos/audit/RiskLimitForm.tsx
```diff
@@ -55,38 +55,37 @@ class RiskLimitForm extends React.Component<FormProps & any, FormState> {
...
-                            max={ toPercent(MAX_RISK_LIMIT) }
-                            minorStepSize={ toPercent(0.001) }
-                            onBlur={ this.onBlur }
-                            stepSize={ toPercent(0.01) }
-                            value={ comparisonField }
-                            onValueChange={ this.onComparisonValueChange } />
-                    </label>
-                </div>
-            </div>
+        const ballotPollingFormField = (
```
dmzimmerman: Didn't they want ballot polling audits eliminated from the UI? Or was that done somewhere else?

ranweiler: Yes, that was removed. I'm holding on to the markup definition and (unobservable) form state in case we want it in the future. We just aren't rendering it right now.

ranweiler: (removed here: 61c46a7e4c5776cbd0844003406240908d4b57ef)

PR #551, opened 2017-08-30
Add Election Information to County Dashboard

### LINKED ISSUE #537: Create endpoint to set election type and date
Once set, ensure we include those data in the DOS and county refresh endpoints.

dmzimmerman:
Closes #537 

Note that the data does _not_ appear on the version of the county dashboard data that is sent to the DoS dashboard, because it already appears on the DoS dashboard anyway.


ranweiler:
> Note that the data does not appear on the version of the county dashboard data that is sent to the DoS dashboard, because it already appears on the DoS dashboard anyway.

Great.

PR #574, opened 2017-08-31
Extreme Upload Acceleration

### LINKED ISSUE #530: Slow upload of CVR file
Per CDOS feedback 8/28:

"A file of 1140 CVRs takes up to a minute to upload. Larger files will take much longer."

dmzimmerman:
This implementation of CVR import parsing can, on my laptop at Galois, import 1,000,000 CVRs in 23 minutes. That compares to... well, the closest known time is @nealmcb's 3+ hours to run out of memory while trying, or @kiniry's 1 hr 40 minutes to import 500,000 with the older code.

Another data point: it takes my machine at home 38 minutes to import 1M CVRs on this branch, as compared with 199 minutes to do so on `master`.

The downside: it uses multiple transactions. There's code to clean up if something goes wrong, but if something goes wrong right at the end, there will be some cruft left in the database. The good part, though, is that the _state_ of the county dashboard will not reflect that CVRs have been uploaded, so when an upload is attempted again, the cruft will be cleaned out.

I think the same technique should probably be applied to ballot manifest imports as well, so I can do that before merging unless anybody has any objections. 

Closes #530 with extreme prejudice.

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
@@ -55,6 +55,11 @@ public class DominionCVRExportParser implements CVRExportParser {
   private static final int BATCH_SIZE = 80;
 
   /**
+   * The size of a batch of CVRs to be committed as a transaction.
+   */
+  private static final int TRANSACTION_SIZE = 400;
+  
+  /**
    * The column containing the CVR number in a Dominion export file.
    */
   private static final int CVR_NUMBER_COLUMN = 0;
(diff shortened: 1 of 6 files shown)


kiniry:
This is great news, @dmzimmerman.  I look forward to taking it for a whirl and evaluating its behavior on the parallel DB server install.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
```diff
@@ -55,6 +55,11 @@
   private static final int BATCH_SIZE = 80;
 
   /**
+   * The size of a batch of CVRs to be committed as a transaction.
+   */
+  private static final int TRANSACTION_SIZE = 400;
```
kiniry: Did you arrive at this number via testing?  Is there a relationship between average CVR size and some cache in the DB?  Something else?  (Magic numbers are scary.)

dmzimmerman: I wanted it to be a multiple of the batch size, and something relatively small, and watching the timing displayed in logs made it seems like a reasonable choice. The reason it's a constant is so that we can tune it later if we need to.

dmzimmerman: (in fact, I could see making it an actual properties setting at some point, along with batch size, so that it could be tuned based on the actual database in use)

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
```diff
@@ -207,6 +216,18 @@ public String endpoint(final Request the_request, final Response the_response) {
       return my_endpoint_result.get();
     }
     
+    // check to be sure that the same county isn't in the middle of a CVR import
+    synchronized (COUNTIES_RUNNING) {
+      if (COUNTIES_RUNNING.contains(county.id())) {
+        transactionFailure(the_response,
```
kiniry: How is this failure handled on the client side?

dmzimmerman: That is an excellent question. I imagine just like any other failure to perform an import.

dmzimmerman: (it should be indistinguishable, to the client, from any other transaction failure)

PR #615, opened 2017-09-01
Fix risk limit step buttons

### LINKED ISSUE #611: Ensure risk limit form state change when up/down buttons used
Keyboard entry always changes both the apparent and internal form state. Currently, it looks like using those buttons does not change the internal form state, but only the external. Investigate, and ensure either entry method can be used.

ranweiler:
Closes #611.


client/src/component/sos/audit/RiskLimitForm.tsx
```diff
@@ -92,6 +93,16 @@ class RiskLimitForm extends React.Component<FormProps & any, FormState> {
         );
     }
 
+    private onButtonClick = (_: number, field: string) => {
```
kiniry: Yuck.  What a style.

PR #637, opened 2017-09-01
Server Modifications from Phase 3 Feedback

### LINKED ISSUE #619: Default order of contests in Contest Update table and on Select Contests page
Let's ask CDOS how they use this table and use that to choose a default order.

From [CDOS 9/1 feedback](https://docs.google.com/document/d/1GpusBIBClPRxmIVRhZrU9qdDAqwsgvvtfdhGDjuDh9Q/edit).

dmzimmerman:
Closes #619


server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
```diff
@@ -255,8 +256,8 @@ private void addContests(final CSVRecord the_choice_line,
       // now that we have all the choices, we can create a Contest object for 
       // this contest (note the empty contest description at the moment, below, 
       // as that's not in the CVR files and may not actually be used)
-      final Contest c = new Contest(cn, my_county.id(), "", choices, 
-                                    the_votes_allowed.get(cn));
+      final Contest c = new Contest(cn, my_county, "", choices, 
+                                    the_votes_allowed.get(cn), contest_count++);
```
ranweiler: I advocate avoiding this C-ism of incrementing-while-passing.

server/eclipse-project/src/main/java/us/freeandfair/corla/json/ContestJsonAdapter.java
```diff
@@ -0,0 +1,178 @@
...
+          break;
+      }
+    }
+    the_reader.endObject();
+    
+    // check if an identically-numbered contest exists
+    
+    if (error || contest_id == null || county == null || description == null ||
+        choices == null || votes_allowed == null) {
+      throw new JsonSyntaxException("invalid data in contest");
```
ranweiler: Would a `JsonParseException` be more appropriate? From the docs:

> This exception is raised if there is a serious issue that occurs during parsing of a Json string. One of the main usages for this class is for the Gson infrastructure. If the incoming Json is bad/malicious, an instance of this exception is raised.

server/eclipse-project/src/main/java/us/freeandfair/corla/json/ContestJsonAdapter.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+    
+    if (sequence_number == null) {
+      sequence_number = 0;
+    }
+    final Contest existing = Persistence.getByID(contest_id, Contest.class);
+    final Contest result;
+    if (existing == null) {
```
ranweiler: At a glance, it seems that if the user encloses a `contest_id`, but it can't be found in the DB, something has gone really wrong, and this error should maybe be unrecoverable. Why should we be permissive here and create a new `Contest`?

server/eclipse-project/src/main/java/us/freeandfair/corla/asm/ASMTransitionFunction.java
```diff
@@ -169,51 +168,69 @@ public ASMTransition value() {
...
+                        SIGN_IN_AUDIT_BOARD_EVENT,
+                        ROUND_IN_PROGRESS)),
+    Q(new ASMTransition(WAITING_FOR_ROUND_SIGN_OFF,
                         SIGN_OUT_AUDIT_BOARD_EVENT,
                         WAITING_FOR_ROUND_SIGN_OFF_NO_AUDIT_BOARD)),
-    M(new ASMTransition(WAITING_FOR_ROUND_SIGN_OFF_NO_AUDIT_BOARD,
+    R(new ASMTransition(WAITING_FOR_ROUND_SIGN_OFF,
+                        ROUND_SIGN_OFF_EVENT,
+                        WAITING_FOR_ROUND_START)),
+    S(new ASMTransition(WAITING_FOR_ROUND_SIGN_OFF,
```
ranweiler: Does this allow a situation where a county X completed a round, county Y completes a round and signs off, hitting the risk limit, and then county X doesn't get a chance to sign off?  Or is the ASM such that we can't achieve the risk limit if some county is in the `WAITING_FOR_ROUND_SIGN_OFF` state?

server/eclipse-project/src/main/java/us/freeandfair/corla/json/ContestJsonAdapter.java
dmzimmerman: Perhaps. We do catch it in most places (it's a parent exception of `JsonSyntaxException`).

server/eclipse-project/src/main/java/us/freeandfair/corla/asm/ASMTransitionFunction.java
dmzimmerman: Remember that there are multiple audit board ASMs, one for each county. The risk limit being achieved is per-county, not per-contest or systemwide.

server/eclipse-project/src/main/java/us/freeandfair/corla/json/ContestJsonAdapter.java
dmzimmerman: Nobody should ever be sending us contests to unserialize - there's no endpoint that accepts them as input. But if they do, there's no good reason why we shouldn't unserialize them, and let any error that might occur when they later want to save the thing to the database occur later... You could imagine, for example, an endpoint that actually lets you specify contest details, in which case you would in fact be creating new objects.

I will, however, readily admit that for _some_ classes - mostly those that we only create internally and that there is no legitimate reason to ever have created externally - I have not been nearly as forgiving in the unserialization phase.

server/eclipse-project/src/main/java/us/freeandfair/corla/asm/ASMTransitionFunction.java
dmzimmerman: (so, the answer to your question is, "no, the situation you describe cannot happen")

Issue #644, opened 2017-09-03
Include audit board ASM state in county refresh


ranweiler:
Right now, the client must fetch this state from an endpoint other than `/county-dashboard`. This means that we can have races where (1) we hit `/audit-board-asm-state`, which indicates that we are in a state with a signed-in audit board, but (2) we don't yet have the actual data for this signed in audit board, because we haven't yet received data from a `/county-dashboard` refresh that includes the audit board. This means that wherever we would _like_ to do things conditionally based on the audit board ASM state, we can't, and need to continue treating derived state as the source of truth, ignoring the ASM.

ranweiler:
Closed in #646.

PR #646, opened 2017-09-03
Audit Board ASM State on County Dashboard

dmzimmerman:
This allows the client to not have race conditions in polling ASM states.


ranweiler:
Closes #644.

PR #647, opened 2017-09-03
CVR and CVRToAudit Query Optimization

### LINKED ISSUE #638: Optimize Performance of cvrsToAuditInRound and cvrsRemainingToAuditInCurrentRound
These two methods in `CountyDashboard` do linear searches across a lazily loaded list, resulting in many, many database queries; this can probably be solved by making an appropriate database query to fetch the records all at once before calling those methods, or eliminating the methods entirely and pushing the code to construct those lists up a level (they are only used by `CountyReport` and `CountyDashboardResponse`). This is not something we want to do every time a `CountyDashboard` is loaded, because it doesn't need all those records most of the time.

dmzimmerman:
Added optimized queries for:

- the random sequence of CVRs to audit (including duplicate samples)
- the list of upcoming ballots to audit (deduplicated and in audit order)
- the list of all ballots audited in a round with all their associated information (deduplicated and in audit order)
- the list of all ballots in a range (including duplicates)

This removes the linear-number-of-query searches through the CVR table for CVRs to audit, the ballots to audit list for duplicates and already-audited ballots during dashboard refreshes (and, thus, the linear number of database queries that went along with them); the lists do obviously still have to be created in the first place when the audit round starts and rendered into JSON for county dashboard refreshes, and audit calculations are done on all ballots in the sequence as they come up - all of those are linear operations, and can't be accelerated beyond minor Java optimizations (if any are still available at this point). But, it _does_ make a constant number of database queries per round started, ballot submitted, and dashboard request, which is a _huge_ improvement.

As an example of the performance increase, I attempted to audit the 1000 ballot sample file to completion (injecting ballot not found every so often so the audit never finishes until the ballots run out) earlier today, and it took 19 minutes; after this query optimization it took 3 3/4 minutes. I ran the 10000 ballot one overnight and it didn't come close to finishing before this query optimization; now it takes about 4 hours (which is still not great, but I have some ideas about that too).

We might want to consider adding an endpoint to get the current list of ballots remaining to audit in the round (in audit order) only if/when the client needs it , rather than including it in every county dashboard refresh (it's actually not clear to me that the client ever really needs it, since it always has the _current_ ballot under audit in the dashboard refresh, as well as round counts and such). The current `/cvrs-to-audit` endpoint actually supplies enough information to be such a list (since it includes audit sequence numbers), it's just sorted in ballot identification order.

Closes #638 with extreme prejudice

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/query/CVRAuditInfoQueries.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/query/CVRAuditInfoQueries.java
@@ -21,6 +21,7 @@ import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
+import javax.persistence.criteria.Subquery;
 
 import org.hibernate.Session;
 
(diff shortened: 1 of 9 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/query/CVRAuditInfoQueries.java
```diff
@@ -81,6 +102,127 @@ private CVRAuditInfoQueries() {
   }
   
   /**
+   * Obtain the sublist of CVRAuditInfo objects for the specified
```
ranweiler: It would be nice if we, in a more human-friendly form, succinctly summarized the intended strategy being used for the queries constructed in these methods.

Issue #653, opened 2017-09-05
Allow SoS user to set date of public meeting


sfsinger19103:
On same page where election date is chosen. 

And prevent user from choosing a random-seed-selection date that is earlier or equal to the election date.

ranweiler:
I've opened #663 for the form validation part of this.

PR #658, opened 2017-09-06
CVR to Audit List Endpoint, and Audit Info, Overhaul

dmzimmerman:
Modifies the `/cvr-to-audit-list` endpoint to take an additional parameter, `include_audited`, that includes already-audited ballots, and adds an `audited` flag to the returned CVR info. Also eliminates the `ballots-to-audit` field from the county dashboard refresh. 

In addition, modifies the server ASM to allow flexible submission (and resubmission) of election type/date, public meeting date (new feature), and risk limit before submitting random seed, and renames/reformats certain dashboard data appropriately.

Addresses #660 and #653 from the server side


server/eclipse-project/src/main/java/us/freeandfair/corla/json/CountyDashboardRefreshResponse.java
```diff
@@ -191,14 +180,9 @@
   private final Round my_current_round;
   
   /**
-   * The election type.
+   * The election info.
```
ranweiler: Should be "The audit info".

Issue #655, opened 2017-09-05
Changes requested in County Update table on SoS page


sfsinger19103:
- Audited contest discrepancies, non-audited contest discrepancies. Plus column for each “reason for audit”.
- A count of the number of ballot cards that have discrepancies in audited contests, and a count of the number of ballot cards that have discrepancies in non-audited contests (noting that some ballot cards will count in both categories)

Per 9/5 conversation with CDOS:
Take out "discrepancy" column.
Add these columns (suggested column titles included, but they might be too long):

- <span>#</span> of ballot cards showing a discrepancy in at least one audited contest "Discrepancies in Contests Under Audit"

For each "reason for audit" (currently "state contest, county contest"), make a column
- <span>#</span> of ballot cards showing a discrepancy in at least one race under audit for the given reason "Discrepancies in State Contests", "Discrepancies in County Contests"

- <span>#</span> of ballot cards showing a discrepancy in at least one non-audited contest "Discrepancies in Contests Not Under Audit"

- <span>#</span> of ballot cards not found "Ballot Cards Not Found"

sfsinger19103:
CDOS has promised (9/5) to sketch the County Update Table that they want and share it with us soon. So there may be more changes in this spec.

dmzimmerman:
There will be no columns for "reasons to audit" without significant rearchitecture of some things, which as far as I can see means it's not happening in the near future. But the rest of this should be addressed soon, server-side.

dmzimmerman:
Which is to say: there will be a column for "discrepancies in audited contests", and a column for "discrepancies in contests not under audit", but not for "audited contests with reason X" and "audited contests with reason Y".

dmzimmerman:
Actually, I want some clarification from them about exactly what they want here. If they want the discrepancies for each category of audited contest (opportunistic contests fall under that list too), there's a good way to do it that feels less "hack"-ish to me than just "let's add yet another field to the dashboard" - namely, replacing the existing "discrepancies" field with a mapping from category to number of discrepancies. But I don't want to get involved in doing that if it's not going to give them what they want.

sfsinger19103:
@dmzimmerman Your immediately previous comment applies to the "Someday" milestone, not the current milestone, right? My current plan is to tell CDOS simply "no" on "audited columns with reason X" and "audited columns for reason Y" for the November 2017 system. Let me know if you want me to change that plan.

If not, are you asking for clarification from them now about what they might want for 2018 and beyond?

[...]

Issue #596, opened 2017-09-01
Support for untabulated ballots: provisionals etc.


nealmcb:
Audits ideally come after all the votes are tabulated, canvassed and reconciled. In Colorado, however, the certification deadline comes shortly after the last date for voters to cure signature verification problems, for provisional ballots to be processed, etc., It is thus possible that a very small number of such ballots may be tabulated after the deadline for CVRs and manifests to be uploaded to the ColoradoRLA tool.  An updated vote count including such ballots may be released at the end of the tabulation and canvass, and the risk limit of the audit needs to apply to the updated outcome.

This is allowed e.g. by Colorado's Rule 25, e.g. for provisional ballots.

RLA theory covers this case. As discussed in [Banuelos-Stark-2012](https://arxiv.org/abs/1207.3413), counties should add a batch of _phantom_ ballot cards to the manifest, which do not correspond to any CVRs, one for each possible late-tabulation ballot card.

To support this, if the total number of ballot cards identified in the manifest is greater than the number of CVRs, the server should augment the sample space for the random selection from the CVRs to match the total from the manifest.  Any selections that don't map to an uploaded CVR should be automatically treated as phantom ballot cards, and handled as in the existing "ballot-not-found" case, i.e. as if votes were marked for all the losers.

sfsinger19103:
What happens currently in the RLA Tool if the Ballot Manifest includes ballots for which there are no corresponding lines in the CVR file?

nealmcb:
I don't think that part of the `corla.pvs` spec has been implemented yet, so the tool doesn't notice the discrepancy at all.

In the converse situation, where the CVR file has lines for ballot ids which are not represented in the ballot manifest, or in particular can't be found among the paper ballots, the Audit Board can click the "Ballot Card Not Found" button, and the ballot is properly treated as a phantom ballot.

So long as the ballot manifest contains at least as many ballots as the CVR file, the audit can properly limit the risk.  But part of the requirements based on Rule 25 are to handle the case of provisionals and other ballots which haven't been counted yet.

dmzimmerman:
It would be straightforward to do a check that all CVRs in the ballot manifest are accounted for in the imported CVR set and generate phantom records for those that are not. We just haven’t done that yet.

dmzimmerman:
[...]

PR #665, opened 2017-09-07
Support Per-Audit-Reason Statistics and Counting Ballots in Ballot Manifests

dmzimmerman:
Addresses #655 from the server side.
Partially addresses #596 from the server side.

Also includes new dashboard info for the ballot count calculated from the ballot manifest and renames the dashboard info for the ballot count calculated from the CVR export.


server/eclipse-project/src/main/java/us/freeandfair/corla/csv/ColoradoBallotManifestParser.java
```diff
@@ -205,6 +211,7 @@ public synchronized boolean parse() {
           break;
         } else {
           my_record_count = my_record_count + 1;
+          my_ballot_count = my_ballot_count + bmi.batchSize();
```
ranweiler: Why do we add `bmi.batchSize()` here?

ranweiler: Oh nevermind, I see, this `+= bmi.batchSize()`. This is a pre-coffee review.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
```diff
@@ -107,7 +107,7 @@ private void updateCountyDashboard(final Response the_response,
       serverError(the_response, "could not locate county dashboard");
     } else {
       cdb.setCVRUploadTimestamp(the_timestamp);
-      cdb.setBallotsCast(the_ballots_cast);
+      cdb.setCVRsImported(the_ballots_cast);
```
ranweiler: :+1:

server/eclipse-project/src/main/java/us/freeandfair/corla/json/CountyDashboardRefreshResponse.java
```diff
@@ -202,10 +208,20 @@
    * @param the_optimistic_ballots_to_audit The optimistic ballots to audit.
    * @param the_ballots_remaining_in_round The ballots remaining in the 
    * current round.
-   * @param the_cast_ballot_count The number of ballots cast.
+   * @param the_ballot_manifest_count The number of ballots represented by the
+   * uploaded ballot manifest.
+   * @param the_cvr_export_count The number of CVRs in the uploaded export file.
    * @param the_audited_ballot_count The number of ballots audited.
+<<<<<<< HEAD
```
ranweiler: Merge conflict artifacts here and below.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Round.java
```diff
@@ -253,45 +258,76 @@ public Integer startAuditedPrefixLength() {
...
+   * Removes a discrepancy for the specified audit reason. This removes it 
+   * both from the total and from the current audit round, if one is ongoing.
+   * 
+   * @param the_reason The reason.
    */
-  public void removeDiscrepancy() {
-    my_discrepancies = my_discrepancies - 1;
+  public void removeDiscrepancy(final AuditReason the_reason) {
+    ensureCounterExists(my_discrepancies, the_reason);
+    my_discrepancies.put(the_reason, my_discrepancies.get(the_reason) - 1);
```
ranweiler: Won't this result in negative discrepancies if `the_reason` was not in `my_discrepancies`, since `ensureCounterExists()` sets the value of `my_discrepancies.get(the_reason)` to 0 in that case?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Round.java
```diff
@@ -253,45 +258,76 @@ public Integer startAuditedPrefixLength() {
...
+   * Adds a disagreement for the specified audit reason. This adds it both to the 
+   * total and to the current audit round, if one is ongoing.
+   * 
+   * @param the_reason The reason.
    */
-  public void addDisagreement() {
-    my_disagreements = my_disagreements + 1;
+  public void addDisagreement(final AuditReason the_reason) {
+    ensureCounterExists(my_disagreements, the_reason);
+    my_disagreements.put(the_reason, my_discrepancies.get(the_reason) + 1);
```
ranweiler: Same concern as above.

server/eclipse-project/src/main/java/us/freeandfair/corla/report/CountyReport.java
```diff
@@ -283,23 +282,32 @@ public Workbook generateExcelWorkbook() {
...
     
     row = summary_sheet.createRow(row_number++);
     cell_number = 0;
     
     cell = row.createCell(cell_number++);
     cell.setCellType(CellType.STRING);
     cell.setCellStyle(bold_style);
+    cell.setCellValue("Total CVRs in CVR Export File");
+    
+    cell = row.createCell(cell_number++);
```
ranweiler: As in the earlier PR, I suggest not using the increment operator on a variable being passed to a method, here, below, and in the other report.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Round.java
dmzimmerman: Yes. Yes it will. Rounds have always been able to have negative discrepancies because if you _resubmit_ a ballot that had previously caused a discrepancy, during a different round than you submitted it, the discrepancy counter still goes down. This is a complete nonissue during regular use because we never resubmit the same ballot twice, and could potentially be confusing if we did... but is correct behavior.

server/eclipse-project/src/main/java/us/freeandfair/corla/report/CountyReport.java
dmzimmerman: In most cases I agree, but here... no. Rewriting both reporting classes to add lines every time I increment the cell number in an Excel file would double the sizes of the classes for no real benefit (there isn't even a readability benefit, because honestly, the code required to generate the Excel files is mostly straight line "make this cell, then make this other cell, then make this row..." anyway).

server/eclipse-project/src/main/java/us/freeandfair/corla/json/CountyDashboardRefreshResponse.java
dmzimmerman: Got 'em.  7fba64b2

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Round.java
dmzimmerman: Actually, this comment helped me see "discrepancies" instead of "disagreements" here, egen though that's not what you pointed out... so thanks! :)

server/eclipse-project/src/main/java/us/freeandfair/corla/model/Round.java
ranweiler: I see. I hadn't realized this because we don't observe it on the client.

PR #666, opened 2017-09-07
Use try-with-resources on download endpoints

dmzimmerman:
Because one instance of this was pointed out in the "security" scan, and because it's generally a good idea, I've converted all the download endpoints to use try-with-resources constructions for their streams.


ranweiler:
🤘

PR #686, opened 2017-09-12
Pairwise Margin Calculation Changes

dmzimmerman:
Previously, we were calculating discrepancies by dealing with sets of winners and losers all at once; while that calculation worked, albeit being occasionally pessimistic, it is harder to have confidence in than a calculation that explicitly computes the discrepancy between each winner and loser in each contest. This introduces such a calculation.

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
@@ -663,27 +663,24 @@ public class CountyContestComparisonAudit implements PersistentEntity, Serializa
     
     // we want to get the maximum pairwise update delta, because that's the "worst"
     // change in a pairwise margin, and the discrepancy we record; we start with
-    // Integer.MAX_VALUE so our maximization algorithm works. it is also the case 
+    // Integer.MIN_VALUE so our maximization algorithm works. it is also the case 
     // that _every_ pairwise margin must be increased for an understatement to be
     // reported
     
     int raw_result = Integer.MIN_VALUE;
     
     boolean possible_understatement = true;
-    boolean discrepancy_found = false;
     
     for (final String winner : my_contest_result.winners()) {
       final int winner_change;
       if (!cvr_choices.contains(winner) && acvr_choices.contains(winner)) {
-        // the winner gained a vote
+        // this winner gained a vote
         winner_change = 1;
-        discrepancy_found = true;
       } else if (cvr_choices.contains(winner) && !acvr_choices.contains(winner)) {
-        // the winner lost a vote
+        // this winner lost a vote
         winner_change = -1;
-        discrepancy_found = true;
       } else {
-        // the winner's votes didn't change
+        // this winner's votes didn't change
         winner_change = 0;
       }
       if (my_contest_result.losers().isEmpty()) {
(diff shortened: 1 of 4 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
```diff
@@ -559,70 +560,64 @@ public Integer computeDiscrepancy(final CastVoteRecord the_cvr,
...
-        result = -1;
+    // now, find the minimum pairwise update delta
+    for (final String winner : my_contest_result.winners()) {
+      final int winner_change;
+      if (!the_cvr_info.choices().contains(winner) && acvr_choices.contains(winner)) {
+        // the winner gained a vote
+        winner_change = 1;
+      } else if (the_cvr_info.choices().contains(winner) && !acvr_choices.contains(winner)) {
+        // the winner lost a vote
+        winner_change = -1;
```
nealmcb: If a winner who was considered earlier in this loop gained a vote or had a winner_change of 0, this would overwrite that result, perhaps resulting in the wrong outcome when compared with https://archive.org/search.php?query=https%3A%2F%2Fwww.stat.berkeley.edu%2F~stark%2FVote%2FauditTools.htm

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
```diff
@@ -559,70 +560,64 @@ public Integer computeDiscrepancy(final CastVoteRecord the_cvr,
...
+      } else {
+        // the winner's votes didn't change
+        winner_change = 0;
```
nealmcb: Similar issue to the comment above.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
dmzimmerman: This would not overwrite that result, because it is factored into the minimization at the end of the second loop. Neither would the one below. However, if it is wrong for some other reason, that would be useful to know.

dmzimmerman: Ah. Yes, I think this is right; because for overstatements it has to widen _every_ pairwise margin, and for understatements it only has to narrow _any_ pairwise margin. I will fix that.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
```diff
@@ -554,72 +628,114 @@ public Integer computeDiscrepancy(final CastVoteRecord the_cvr,
...
+   * otherwise.
+   */
+  @SuppressWarnings({"PMD.ModifiedCyclomaticComplexity", "PMD.StdCyclomaticComplexity",
+                     "PMD.NPathComplexity"})
+  private OptionalInt computeAuditedBallotDiscrepancy(final CVRContestInfo the_cvr_info,
+                                                      final CVRContestInfo the_acvr_info) {
+    // we want to get the maximum pairwise update delta, because that's the "worst"
     // change in a pairwise margin, and the discrepancy we record; we start with
-    // Integer.MAX_VALUE so our minimization algorithm works
+    // Integer.MAX_VALUE so our maximization algorithm works. it is also the case
```
ranweiler: This comment is now out of date.

PR #702, opened 2017-09-14
Client sprint

ranweiler:


ranweiler:
On Linux, to test the file upload spinners in dev mode, you can simulate latency on the loopback interface via:
```
sudo tc qdisc add dev lo root netem delay 50ms
```
And disable it with:
```
sudo tc qdisc del dev lo root netem
```
I recommend only injecting latency with the first command right before you attempt a file upload, or it will be annoying to wait on the artificially-delayed dev server chatter.

client/src/adapter/dosDashboardRefresh.ts
```diff
@@ -25,10 +25,26 @@ function parseRounds(rounds: any[]) {
...
-function parseDisCount(data: any): number {
+function parseDisagreementCount(data: any): number {
+    if (_.isEmpty(data)) {
+        return null;
+    }
+
     return _.sum(_.values(data));
 }
 
+function parseDiscrepancyCounts(data: any): any {
```
dmzimmerman: I see that these had to be split into their own functions. Pity about the loss of the old function name. :)

PR #711, opened 2017-09-14
Prevent Round 0

### LINKED ISSUE #700: Round 0 in Progress?!
Not sure how I got here:

(Edit by @ranweiler) Expand details below for image (warning: flashing screen):
<details>

round 0

</details>

dmzimmerman:
Previously, in the very unlikely situation where an audit was started and no counties actually had to audit anything (because they had already achieved their risk limits, or missed their deadlines, or had no selected driving contests), the state dashboard would incorrectly display a Round 0, because it didn't know the audit was done. Now it does.

Closes #700


dmzimmerman:
I have actually verified, by running the client and uploading CVRs and a ballot manifest for a county, then not choosing any contests to audit, that this works correctly.

PR #713, opened 2017-09-14
Add Contest Discrepancies to DoS Dashboard, Make Logging More Sane

dmzimmerman:
This adds contest discrepancy information to the DoS dashboard refresh, and also reduces our syslog output by switching some endpoints that are frequently used and read-only to DEBUG level instead of INFO level.

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/json/DoSDashboardRefreshResponse.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/json/DoSDashboardRefreshResponse.java
@@ -163,8 +163,10 @@ public class DoSDashboardRefreshResponse {
                 Math.max(estimated, 
                          Math.max(0, ccca.estimatedBallotsToAudit() - 
                                      ccca.dashboard().auditedPrefixLength()));
+            
+            // possible discrepancy types range from -2 to 2 inclusive,
+            // and we provide them all in the refresh response
             for (int i = -2; i <= 2; i++) {
-              // get the discrepancy counts of various types
               if (discrepancy.get(i) == null) {
                 discrepancy.put(i,  0);
               }
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/json/DoSDashboardRefreshResponse.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/json/DoSDashboardRefreshResponse.java
@@ -11,11 +11,13 @@
 
 package us.freeandfair.corla.json;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
+import java.util.SortedMap;
+import java.util.TreeMap;
 
 import javax.persistence.PersistenceException;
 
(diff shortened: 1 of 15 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/json/DoSDashboardRefreshResponse.java
```diff
@@ -120,22 +135,45 @@ protected DoSDashboardRefreshResponse(final ASMState the_asm_state,
...
+                         Math.max(0, ccca.optimisticBallotsToAudit() - 
+                                     ccca.dashboard().auditedPrefixLength()));
+            estimated = 
+                Math.max(estimated, 
+                         Math.max(0, ccca.estimatedBallotsToAudit() - 
+                                     ccca.dashboard().auditedPrefixLength()));
+            for (int i = -2; i <= 2; i++) {
+              // get the discrepancy counts of various types
+              if (discrepancy.get(i) == null) {
+                discrepancy.put(i,  0);
```
ranweiler: Extreme nit: extra space.

server/eclipse-project/src/main/java/us/freeandfair/corla/json/DoSDashboardRefreshResponse.java
```diff
@@ -120,22 +135,45 @@ protected DoSDashboardRefreshResponse(final ASMState the_asm_state,
...
+               CountyContestComparisonAuditQueries.matching(cta.contest())) {
+            optimistic = 
+                Math.max(optimistic, 
+                         Math.max(0, ccca.optimisticBallotsToAudit() - 
+                                     ccca.dashboard().auditedPrefixLength()));
+            estimated = 
+                Math.max(estimated, 
+                         Math.max(0, ccca.estimatedBallotsToAudit() - 
+                                     ccca.dashboard().auditedPrefixLength()));
+            for (int i = -2; i <= 2; i++) {
```
ranweiler: Maybe add a comment here to note that these are the possible discrepancy values. It was obvious after recent code review, but since it isn't encoded in an enum, might be worth remarking.

dmzimmerman: I have been meaning to encode it in an enum. Maybe now is the right time to do that.

nealmcb:
I am THRILLED to see this - just the detail we need: `contest_id` and all the data needed for verifying the RLA calculations themselves! Thank you @dmzimmerman!!
```  "discrepancy_count": {
    "149": {
      "-1": 0,
      "0": 0,
      "-2": 0,
      "1": 0,
      "2": 1
    }```

nealmcb:
I like the selection of log messages, mostly GETs, to not bother logging. That will make the default logging much more readable.

server/eclipse-project/src/main/java/us/freeandfair/corla/json/DoSDashboardRefreshResponse.java
dmzimmerman: Or, maybe after this build is delivered to CDOS and I have a bit more time to do the right thing...

sfsinger19103:
Does this change what CDOS sees? If so, has CDOS OK'd the change?

PR #746, opened 2017-09-21
Docs post p2

sfsinger19103:
Substantial improvement of user documentation. Still needs post-processing by hand of user documentation files.


dmzimmerman:
It's possible that the footer logos are already fixed on the files we're actually sending in (the ones with the nice breaks), since I fixed them last week. If that's the case, we can leave the others as is for now, but it would be good to have them all be consistent.

sfsinger19103:
These are actually v.1.0.0.alpha run books. beta version run books will be completed over the weekend, based on finalized beta system. So it would be wrong to change the version number.

Moved the logo up in the footer. Hope you like it, @dmzimmerman ! And thanks for taking the time on this pull request.

PR #777, opened 2017-09-27
Fixed a parsing error for number of candidates to vote for in a contest.

dmzimmerman:
Self-explanatory

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
@@ -253,8 +253,9 @@ public class DominionCVRExportParser implements CVRExportParser {
         count = count + 1;
       }
       // get the "(Vote For=" number from the contest name and clean up the name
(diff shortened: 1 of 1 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
```diff
@@ -253,8 +253,9 @@ private void updateContestStructures(final CSVRecord the_line,
         count = count + 1;
       }
       // get the "(Vote For=" number from the contest name and clean up the name
```
ranweiler: This comment is now stale, because we "clean up the name" (`trim()`) _later_, to squash a bug. Maybe add a comment to the place where the trimming now occurs that indicates it must happen after the `replace(cn, "")` which is used to build `vf`.

dmzimmerman: I disagree with this assessment; the comment still accurately reflects what the 3 lines below are doing. However, it is clearer when split into two places, so I have done so.

PR #778, opened 2017-09-28
Tie Contests and Per-Contest-Per-Record Discrepancy/Disagreement Recording

### LINKED ISSUE #773: Graceful server behavior in case of tied contest
The RLA Tool is not required to support the audit of a contest for which the CVRs indicate an exact tie. 

However, should that case arise, the RLA Tool should behave gracefully and give useful information to the user. 

Operationally, the RLA Tool will have sufficient information when a CVR file with a tied contest is uploaded. (Note that, although Rule 25 asks the Secretary of State to choose contests for audit on the Friday after Election Day, which is before the Counties will have uploaded CVR files, the SoS can identify a contest in the RLA Tool only after the CVR files for that contest has been uploaded by the County.

How should this be implemented? Here's an idea to start the discussion:
 
> as part of the processing of any uploaded CVR file, change the status of any tied contest automatically to "hand count required"?

dmzimmerman:
Server changes for the final deliverable. 

- tied contests are detected and properly flagged
- per-contest per-CVR/ACVR-pair discrepancy and disagreement information is tracked and available in the database in the `county_contest_comparison_audit_discrepancy` and `county_contest_comparison_audit_disagreement` tables.

Closes #773


nealmcb:
Thank you thank you, Dan!
My initial tests suggest the new detailed discrepancy tables (with 1-vote overstatements, 2-vote understatements and the like) are working beautifully!

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
```diff
@@ -197,32 +204,38 @@
    * The number of two-vote understatements recorded so far.
    */
   @Column(nullable = false)
-  private Integer my_two_vote_under = 0;
+  private Integer my_two_vote_under_count = 0;
```
ranweiler: Does this field name change also change the serialization?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestResult.java
```diff
@@ -507,18 +507,24 @@ public void updateResults() {
...
         sorted_totals.entrySet().iterator();
     Entry<Integer, List<String>> entry = null;
     int votes = 0;
-    while (vote_total_iterator.hasNext() && my_winners.size() < my_votes_allowed) {
+    while (vote_total_iterator.hasNext() && my_winners.size() < my_winners_allowed) {
       entry = vote_total_iterator.next();
       votes = entry.getKey();
-      my_winners.addAll(entry.getValue());
+      final List<String> choices = entry.getValue();
+      if (choices.size() + my_winners.size() < my_winners_allowed) {
```
ranweiler: Should this be `<=`? We do want to add this loop iteration's `choices` if adding them to `winners` causes the new `winners` size to _exactly_ match `my_winners_allowed`, right?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
```diff
@@ -199,26 +206,55 @@ private void parseFile(final Response the_response, final UploadedFile the_file)
...
+   * @return the number of tied contests detected.
+   */
+  private int handleTies(final County the_county) {
+    int result = 0;
+    final Set<CountyContestResult> contest_results = 
+        CountyContestResultQueries.forCounty(the_county);
+    final DoSDashboard dosdb = Persistence.getByID(DoSDashboard.ID, DoSDashboard.class);
+    
+    for (final CountyContestResult ccr : contest_results) {
+      if (ccr.minMargin() == 0) {
```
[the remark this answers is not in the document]
nealmcb: Should be the smallest pairwise margin among all pairs of (reported winner, reported loser). I.e. the winner with the fewest votes and the loser with the most.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
```diff
@@ -306,8 +306,12 @@ private void addContests(final CSVRecord the_choice_line,
       // now that we have all the choices, we can create a Contest object for 
       // this contest (note the empty contest description at the moment, below, 
       // as that's not in the CVR files and may not actually be used)
+      // note that we're using the "Vote For" number as the number of winners
+      // allowed as well, because the Dominion format doesn't give us that
+      // separately
```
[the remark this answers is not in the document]
nealmcb: We know this is true for Colorado, but not for e.g. Pennsylvania, since there are voting methods in use in Pennsylvania that we don't handle. And perhaps note that this won't work (none of it will!) with IRV etc.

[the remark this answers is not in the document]
nealmcb: And yes, this has been discussed with CDOS....

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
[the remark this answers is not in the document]
dmzimmerman: Yes, that’s correct. It is (supposed to be) the minimum margin between any winner and any loser (recall that contests can have more than one winner).

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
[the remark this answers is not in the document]
dmzimmerman: It’s not lack of documentation, it’s lack of information - the dominion csv format simply doesn’t distinguish between these things.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
dmzimmerman: These objects are not serialized directly. You should not notice it. It does change database column names.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestResult.java
dmzimmerman: I’ll look into that.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
ranweiler: Great.

server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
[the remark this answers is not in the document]
ranweiler: Got it, so we can't ever know get that info from their format alone, so for now we just have to assume _something_.

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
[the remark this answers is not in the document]
ranweiler: Right, good.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestResult.java
dmzimmerman: Not only should it have been, but there was a problem in the code below it too.

dmzimmerman: Addressed in 86abdb60

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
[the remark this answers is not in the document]
dmzimmerman: Looking at `calculateMargins` reveals:

```
  private void calculateMargins() {
    my_min_margin = Integer.MAX_VALUE;
    my_max_margin = Integer.MIN_VALUE;
    for (final String w : my_winners) {
      if (my_losers.isEmpty()) {
        my_min_margin = 0;
        my_max_margin = 0;
      } else {
        for (final String l : my_losers) {
          final int margin = my_vote_totals.get(w) - my_vote_totals.get(l);
          my_min_margin = Math.min(my_min_margin, margin);
          my_max_margin = Math.max(my_max_margin, margin);
        }
      }
    }
  }
```

So yes, it is what I said it was.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestResult.java
```diff
@@ -527,14 +525,8 @@ public void updateResults() {
...
     while (vote_total_iterator.hasNext()) {
-      // all the other choices that have the same number of votes as the last
-      // winner count as winners
-      entry = vote_total_iterator.next();
-      if (entry.getKey() == votes) {
-        my_winners.addAll(entry.getValue());
-      } else {
-        my_losers.addAll(entry.getValue());
-      }
+      // all the other choices count as losers
```
ranweiler: Checking my understanding: the old code was wrong because, if we get into this loop, we have previously broken the previous `while` loop because `my_winners.size() >= my_winners_allowed` (if we broke out of the previous `while` loop because we exhausted the iterator, we'll never enter this one). That means that found as many winners as we are allowed to find. Thus, there's no way (no matter how many votes were earned) anything else could be a winner.

That would make sense, except what if there are multiple ties for various places? Then what we call "winners" are just determined by the iteration order, right? For concreteness, suppose we have a contest with 2 winners allowed. The votes are:
- `A` has 10 votes
- `B` has 5 votes
- `C` has 5 votes
- `D` has 2 votes

Then we'd call `A` and `B` winners, stop iterating, and call `C` and `D` losers, even though `B` and `C` are tied. Does some other precondition prevent us from reaching this scenario?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/DoSDashboard.java
```diff
@@ -206,25 +206,27 @@ public boolean removeContestsToAuditForCounty(final County the_county) {
...
+        contest_to_remove = c;
+        break;
       }
     }
-    my_contests_to_audit.removeAll(contests_to_remove);
-    if (auditable && the_contest_to_audit.audit() != AuditType.NONE) {
-      my_contests_to_audit.add(the_contest_to_audit);
+    
+    if (auditable) {
+      my_contests_to_audit.remove(contest_to_remove);
```
ranweiler: Suppose `the_contest_to_audit` has audit type `NOT_AUDITABLE`. Do we expect this to end up in `my_contests_to_audit`?

ranweiler: Can a contest in `my_contests_to_audit` _ever_ have a contest type of `NOT_AUDITABLE`? If not, should we really be checking `if (!the_contest_to_audit.audit().equals(AuditType.NOT_AUDITABLE))`?

dmzimmerman: Yes, it can have type `NOT_AUDITABLE`; if it does, it was placed there at CVR import time, and can only be removed from the list if the contest itself goes away (by virtue of the CVR import being overwritten by a new one), using the method `removeContestsToAuditForCounty`.

dmzimmerman: Remember that the audit type of the parameter to that method is the _desired_ audit type for the list (so, to change a contest from `COMPARISON` to `HAND_COUNT`, you pass in the same contest in a `ContestToAudit` object with audit type `HAND_COUNT`).

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestResult.java
dmzimmerman: We _would_, in that case, call `C` and `D` losers, but we'd know there was a tie, the minimum margin would be 0, and the contest would be unauditable. It is, essentially irrelevant which members of a tied group are called "winners" and which are called "losers", because any time a tie exists, the contest is not auditable.

dmzimmerman: (unless the tie is between losers, of course)

server/eclipse-project/src/main/java/us/freeandfair/corla/model/DoSDashboard.java
ranweiler: @dmzimmerman that makes sense. I'm not sure what the answer is to my very first question: if `the_contest_to_audit` has audit type `NOT_AUDITABLE` (which, like you said, is what we _want_ its audit type to _become_), do we expect it to end up in `my_contests_to_audit`?

dmzimmerman: Yes, it always ends up there at CVR import time.

dmzimmerman: When the CVR import process calls this method with a type of `NOT_AUDITABLE`, the contest does in fact end up in the list.

ranweiler: Okay, so suppose a contest `c` is `equal` to `the_contest_to_audit` and exists in `my_contests_to_audit`. Either it is auditable or it is not (i.e. has audit type `NOT_AUDITABLE`).

If it is auditable, then we will remove `c` from `my_contests_to_audit`. Then, as long as `the_` _doesn't_ have audit type `NONE`, it will be added to `my_contests_to_audit`, either of which is an "update".

If `c` is _not_ auditable, then it has audit type `NOT_AUDITABLE`. That means we won't enter the true branch in `if (auditable) { ...`, and so we will neither remove `c` nor add `the_`. That means, in particular, we can never set a `NOT_AUDITABLE` contest to have any other type, which was the goal. It also means we can't use this method to update any other data which `equals` is insensitive to, so I guess we just need to be sure that is fine.

dmzimmerman: Note that the _equals()_ is comparing the _contests_ in the `ContestToAudit` objects, not the `ContestToAudit` objects themselves. So I don't think there is any other data to consider.

ranweiler: @dmzimmerman right! Okay, so, like I said, that means if we have a `contest` that is ever set to `NOT_AUDITABLE`, we can't, for example, ever update the `reason` field of the associated `ContestToAudit` in `my_contests_to_audit`, but I'm pretty sure we don't care about that at all.

PR #779, opened 2017-09-29
Use config to define default notice timeout

ranweiler:


client/src/notice.ts
```diff
@@ -1,29 +1,28 @@
 import { Intent, Position, Toaster } from '@blueprintjs/core';
 
-
-const Toast = Toaster.create({ autoFocus: true });
+import { defaultNoticeTimeout } from 'corla/config';
 
 
-const DEFAULT_TIMEOUT = 5000;
+const Toast = Toaster.create({ autoFocus: true });
```
dmzimmerman: Toaster?

PR #780, opened 2017-09-29
Individual Audit Statuses, and Some Renaming

dmzimmerman:
Adds individual audit statuses to the `county_contest_comparison_audit` table: `NOT_STARTED`, `IN_PROGRESS`, `RISK_LIMIT_ACHIEVED`, `ENDED`. `ENDED` means "ended without achieving the risk limit" (either because it was aborted and sent to hand count, or aborted for some other reason - currently not possible - or because it was an opportunistic contest and the county's audit ended). 

This branch also renames "ballots" to "samples" everywhere it is appropriate to do so in the Java code (assuming I caught everything), but does _not_ change the JSON exposed to the client (it still says "ballots"), so it does not require any client changes.


server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
```diff
@@ -486,11 +528,35 @@ private BigDecimal computeOptimisticBallotsToAudit(final int the_two_under,
...
+   * 
+   * @param the_count The count of samples that have been unaudited simultaneously
+   * (for duplicates).
+   */
+  public void signalSampleUnaudited(final int the_count) {
+    my_estimated_recalculate_needed = true;
+    my_audited_sample_count = my_audited_sample_count - the_count;
+    
+    if (my_audit_status != AuditStatus.ENDED) {
+      my_audit_status = AuditStatus.IN_PROGRESS;
```
ranweiler: With this, starting with no audited samples, if we did:
```java
signalSampleAudited(1);
signalSampleUnaudited(1);
```
That leave us in a state where the audit status is `IN_PROGRESS` but `my_audited_sample_count` is 0. Should we prevent that? Does it matter, in context?

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
```diff
@@ -887,7 +887,27 @@ public Integer auditedSampleCount() {
...
+    }
+  }
+  
+  /**
+   * Updates the status for all audits in the county. This changes their states
+   * based on whether they have achieved their risk limits.
+   */
+  public void updateAuditStatus() {
+    for (final CountyContestComparisonAudit ccca : my_comparison_audits) {
+      ccca.updateAuditStatus();
```
ranweiler: If one of these calls failed, would all the transactions roll back, or not? Same question applies above.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
dmzimmerman: I suspect it doesn't matter in context... the audit really is "in progress", because you've started it and haven't stopped it, and in actual usage `signalSampleAudited` would be called again right after `signalSampleUnaudited`.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyDashboard.java
dmzimmerman: The entire transaction would roll back. However, I don't see any realistic way that those calls can fail while simultaneously having the actions that preceded them succeed.

ranweiler: Great.

server/eclipse-project/src/main/java/us/freeandfair/corla/model/CountyContestComparisonAudit.java
ranweiler: Yeah, that part makes sense. I guess we'll just have to be careful not to incorrectly assume that "in progress" implies "at least one sample audited".

dmzimmerman: Given that the sample count is actually part of the record, I suspect that won't be a problem. This is likely to be only used in reporting anyway.

dmzimmerman:
I added another status `NOT_AUDITABLE`, which is assigned to contests that are tied.

PR #799, opened 2017-10-02
Client tied contests

### LINKED ISSUE #774: Graceful client behavior in case of a tied contest
See issue #773  for the server side. 

How should the client help the SoS avoid choosing a tied contest for audit in the RLA Tool? Should the sos/audit/select-contests page have a contest status column (which might be automatically set by the server to "hand count" in case of a tie)?  Or should the checkbox for a tied contest be disabled, with some kind of message by the checkbox? Or should checking the box for a tied contest throw an error message (e.g., "this contest is tied and cannot be used to determine a risk level")?

ranweiler:
Closes #774.


ranweiler:
Example:
screenshot from 2017-10-02 14-06-53

Issue #788, opened 2017-10-02
buggy behavior of BM upload during long CVR upload


sfsinger19103:
If the CVR file is uploaded first and while the circle is spinning it is possible to browse to the ballot manifest file, enter the hash value and click on the Upload button for the ballot manifest. The status for the ballot manifest file will falsely indicate that the ballot manifest file has been uploaded, but when the CVR file is finished being uploaded and the circle goes away the ballot manifest status reverts to not uploaded. The Upload and Re-upload buttons should be deactivated while an upload is being performed, especially for the very long upload times that the counties will see for the larger CVR files

from RLA Tool Feedback #1_tool version3.docx

ranweiler:
I did some investigating, and was able to repro this consistently (using `tc` to simulate high upload latency/big file uploads).

1. Set latency.
1. Select both the ballot manifest and CVR export, enter the hashes, but do not click either upload button.
1. Click the CVR upload button.
1. Click the manifest upload button. 

The client *appears to be working as expected*. The reason it seems *not* to be is that, in the dashboard refresh JSON, I am getting both an uploaded ballot manifest (a non-empty `ballot_manifest_file` key) and an uploaded CVR export (a non-empty `cvr_export_file`), but the `asm_state` is `CVRS_OK`, which is wrong. Either the ASM state should be `BALLOT_MANIFEST_AND_CVRS_OK`, or the ballot manifest key should have no value. We're using the ASM state to decide if the ballot manifest has been uploaded (not the presence of some data), and so we observe an apparently empty form. We don't want to fix the display issue by checking for the data instead of the ASM, because then we'd be left in a state where it _looks_ (to the county user) like they uploaded all the necessary files, but to the DOS, it will look like they haven't imported a ballot manifest.

I suspect an ASM bug triggered by interleaved two-step uploads.

One trace of events that triggers the bug is as follows (as observed via client notifications):
1. CVR explort _uploaded_
1. Ballot manifest _uploaded_
1. Ballot manifest _imported_
1. CVR export _imported_

Relevant server logs:
```
21:02:20,854  INFO corla:399 - endpoint /upload-file hit by localhost:8888
21:02:20,854  INFO corla:147 - handling file upload request from 0:0:0:0:0:0:0:1
21:02:20,855  INFO corla:172 - successfully saved file of size 6187 from 0:0:0:0:0:0:0:1
21:02:21,805  INFO corla:399 - endpoint /contest/county hit by localhost:8888
[...]

PR #804, opened 2017-10-03
Attempt to fix ASM transitions for multi-transaction endpoints.

dmzimmerman:
Nearly all endpoints are transactional; CVR imports, however, are multi-transaction. As a result, if somebody imports a ballot manifest while a CVR is being imported, it may not be noticed. This is an attempt to address that.


ranweiler:
Closes #788.

PR #811, opened 2017-10-05
Better error handling for bad CVR import files.

dmzimmerman:
Also eliminates the last bastion of transactions being used manually in query classes.

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/persistence/Persistence.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/persistence/Persistence.java
@@ -300,7 +300,19 @@ public final class Persistence {
       }
     }
   }
-
+  
+  /**
+   * @return true if a session is open on this thread, false otherwise.
+   * @exception IllegalStateException if the database isn't running.
+   */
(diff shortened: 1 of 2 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/persistence/Persistence.java
```diff
@@ -359,11 +371,16 @@ public static boolean beginTransaction()
...
     boolean result = true;
-    final Session session = currentSession();
+    Session session = currentSession();
+    
     if (isTransactionActive()) {
       result = false;
     } else if (canTransactionRollback()) {
       rollbackTransaction();
+      // session is explicitly closed by rollback
+      // interesting note: isOpen() on the session _still returns true_ even though
```
ranweiler: Does `isSessionOpen()` return `true` in this case, too?

dmzimmerman: Yes, it does, but I still think it's a worthwhile thing to have for potential other uses. I think I should also report it to the Hibernate people, when I get a chance.

PR #821, opened 2017-10-06
Count tied contests as unaudited when computing discrepancies

### LINKED ISSUE #819: Count Tied Contests as Unaudited (Client Side)
In 1.0.0, discrepancies and disagreements in tied contests are included in the totals for "audited" contests; they should be included in the totals for "unaudited" contests.

ranweiler:
Closes #819.

### DIFF
--- a/client/src/adapter/dosDashboardRefresh.ts
+++ b/client/src/adapter/dosDashboardRefresh.ts
@@ -40,7 +40,8 @@ function parseDiscrepancyCounts(data: any): any {
 
     const total = _.sum(_.values(data));
     const opportunistic = data.OPPORTUNISTIC_BENEFITS || 0;
-    const audited = total - opportunistic;
+    const tied = data.TIED_CONTEST || 0;
+    const audited = total - opportunistic - tied;
 
     return { audited, opportunistic };
 }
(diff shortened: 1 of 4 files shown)


client/src/adapter/dosDashboardRefresh.ts
```diff
@@ -40,7 +40,8 @@ function parseDiscrepancyCounts(data: any): any {
 
     const total = _.sum(_.values(data));
     const opportunistic = data.OPPORTUNISTIC_BENEFITS || 0;
-    const audited = total - opportunistic;
+    const tied = data.TIED_CONTEST || 0;
+    const audited = total - opportunistic - tied;
 
     return { audited, opportunistic };
```
dmzimmerman: Shouldn't `return { audited, opportunistic }` be `return { audited, opportunistic + tied }`?

ranweiler: Maybe. I was thinking we didn't care about tied contests _at all_, since they can't be audited by the current system. But I should make it consistent with whatever you're doing in the reports.

dmzimmerman: I think it makes sense to count them _somewhere_, because a discrepancy is still a discrepancy - they are supposed to record all the choices on the ballots they audit.

PR #826, opened 2017-10-10
Fix a concurrent upload issue and some error handling

dmzimmerman:
Fixes an issue where concurrently-uploaded CVR sets that contained tied contests could trigger transaction failures (which, technically, should not have been a problem - but would have manifested as failed uploads when we could have prevented it from doing so). Also more gracefully handles requests for lists of CVRs to download that are inappropriate (for counties with no uploaded CVRs, or for inappropriate rounds, or other issues).

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
@@ -58,6 +58,16 @@ import us.freeandfair.corla.util.UploadedFileStreamer;
     "PMD.StdCyclomaticComplexity"})
 public class CVRExportImport extends AbstractCountyDashboardEndpoint {
   /**
+   * The number of times to retry a DoS dashboard update operation.
+   */
+  private static final int DOS_DASHBOARD_UPDATE_RETRIES = 10;
+  
+  /**
+   * The number of milliseconds to sleep between transaction retries.
+   */
+  private static final long TRANSACTION_SLEEP_MSEC = 1000;
+  
+  /**
    * The " (id " string.
    */
   private static final String PAREN_ID = " (id ";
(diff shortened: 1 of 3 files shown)


server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
```diff
@@ -206,55 +217,150 @@ private void parseFile(final Response the_response, final UploadedFile the_file)
...
+   * Calculates a delay, in milliseconds, to sleep before retrying a transaction.
+   * 
+   * @param the_county The county to retry for.
+   * @param the_retries The number of retries so far.
+   */
+  private long calculateTransactionSleep(final County the_county, final int the_retries) {
+    final int county_id_mod = 4;
+    final int retry_fraction = 2;
+    
+    return (the_county.id() % county_id_mod + the_retries / retry_fraction) *
```
ranweiler: I take it that the `%` is just to add jitter to the backoff?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CountyReportDownload.java
```diff
@@ -66,9 +98,22 @@ public AuthorizationType requiredAuthorization() {
...
+    County county = Main.authentication().authenticatedCounty(the_request);
+
+    if (county == null) {
+      county = 
+          Persistence.getByID(Long.parseLong(the_request.queryParams(COUNTY)), County.class);
+      if (county == null) {
+        badDataContents(the_response, "county " + the_request.queryParams(COUNTY) +
+                                      " does not exist");
+      }
+      assert county != null; // makes FindBugs happy
```
ranweiler: Do we really want to have an assertion failure here, if a DOS user doesn't specify a valid county ID? Will this interrupt the HTTP response?

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
dmzimmerman: Yup

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CountyReportDownload.java
ranweiler: Answer: we will never reach that code, because `badDataContests()` calls `halt()`.

dmzimmerman: badDataContents halts first in any situation where the assertion would fail.

PR #827, opened 2017-10-11
Updates to support CVR import progress

dmzimmerman:
This is an attempt to add progress indication to long-running CVR imports, from the server side.

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/csv/DominionCVRExportParser.java
@@ -38,6 +38,7 @@ import us.freeandfair.corla.model.Choice;
 import us.freeandfair.corla.model.Contest;
 import us.freeandfair.corla.model.County;
 import us.freeandfair.corla.model.CountyContestResult;
+import us.freeandfair.corla.model.CountyDashboard;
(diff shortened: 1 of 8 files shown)


ranweiler:
The response to my (only) request to the `import-cvr-export` endpoint has body:
```
{
  "result": "county 1 is already importing CVRs, try later"
}
```

ranweiler:
Also: the `approximate_record_count` is not appearing the `UploadedFile` JSON serialization because it has a custom JSON adapter which needs to be updated, too.

dmzimmerman:
Well, I should have known it wouldn't quite work on the first, rushed try. :) 

`approximate_record_count` in JSON addressed in 020c85ff

dmzimmerman:
Interestingly, the re-upload failure is almost certainly something that would have happened _before_, as well, in the event of contests with ties...

dmzimmerman:
Yup. It's definitely something that would have happened before; we just hadn't tested it. I was removing the contest records before removing the tied contest "contest to audit" records in the DoS dashboard. Addressed in 170cebec.

IMPORTANT NOTE: if the ballot manifest is re-uploaded while CVRs are still being imported, the CVR import _will fail_ (properly, because the ballot manifest has changed from underneath it). Thus, it is probably a good idea to disable the ballot manifest upload button while CVRs are being imported.

PR #839, opened 2017-10-21
Server Fixes based on CDOS and Other Reports

### LINKED ISSUE #836: rla_export says ERROR:root:get_endpoint: Status code 500 for http://localhost:8888/state-report
If the new `rla_export` tool is run very early in an audit, e.g. perhaps before the random seed is drawn, it can produce this message when trying to download the `state_report.xlsx`:

`ERROR:root:get_endpoint: Status code 500 for http://localhost:8888/state-report`

Note that this can't happen with the GUI UI, since the user has no button to ask for the audit report until it is ready. There is not much to report on before the seed is drawn, so just having some sort of "not ready yet" error would be fine.

The 500 error is related to this error in the server logs.

```
09:32:01,641 ERROR GeneralError:54 - 
java.lang.NullPointerException: instant
  at java.util.Objects.requireNonNull(Objects.java:228)
 ...
```

### LINKED ISSUE #838: File status in "uploaded_file" table is not always updated
As noted by CDOS during test runs, the `uploaded_file` table is not always updated properly. This is because, when transactions are retried due to conflicts with the DoS dashboard information, the update to the `uploaded_file` table  is thrown away.

### LINKED ISSUE #840: Database transactions hang when Spark threads die
If a Spark thread dies due to an unhandled exception, the code in `AbstractEndpoint` that logs the endpoint hit and result and commits the database transaction is never executed. This means that the database transaction stays open, and since we manually manage Hibernate sessions, it never closes unless the database decides to kill it for some reason... and that can, more often than not, result in the hanging of other transactions due to conflicts. 

We can mitigate this by making sure that Spark threads never die untimely deaths, or at least by doing as much as we can to make sure of that.

### LINKED ISSUE #841: Version Number UNKNOWN
When running from the shaded Jar, under most circumstances, the version number in the log displays as UNKNOWN.

dmzimmerman:
Fixes the issue where a 500 error is generated and server hangs when a report is generated at an inappropriate time.

Fixes the issue noted by CDOS where the "uploaded_file" table status does not get updated

Fixes UNKNOWN version numbers.

Attempts to ensure that database transactions are not left hanging, by keeping Spark threads from dying unexpectedly in unanticipated, but not server life threatening, situations.

Closes #836 
Closes #838
Closes #840 
Closes #841

### DIFF
--- a/server/eclipse-project/pom.xml
+++ b/server/eclipse-project/pom.xml
@@ -203,5 +203,10 @@
 			<artifactId>maven-model</artifactId>
 			<version>3.5.0</version>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
+			<version>3.6</version>
+		</dependency>
 	</dependencies>
 </project>
--- a/server/eclipse-project/pom.xml
+++ b/server/eclipse-project/pom.xml
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>us.freeandfair.production</groupId>
 	<artifactId>colorado_rla</artifactId>
-	<version>1.0.4</version>
+	<version>1.0.4-dev</version>
 	<name>ColoradoRLA</name>
 	<description>A risk-limiting audit system for the State of Colorado</description>
 	<properties>
--- a/server/eclipse-project/pom.xml
+++ b/server/eclipse-project/pom.xml
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>us.freeandfair.production</groupId>
 	<artifactId>colorado_rla</artifactId>
-	<version>1.0.3</version>
+	<version>1.0.4</version>
 	<name>ColoradoRLA</name>
 	<description>A risk-limiting audit system for the State of Colorado</description>
 	<properties>
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
@@ -384,7 +384,7 @@ public class CVRExportImport extends AbstractCountyDashboardEndpoint {
    */
   @Override
   @SuppressWarnings({"PMD.ConfusingTernary"})
-  public String endpoint(final Request the_request, final Response the_response) {    
+  public String endpointBody(final Request the_request, final Response the_response) {    
     // we know we have county authorization, so let's find out which county
     final County county = Main.authentication().authenticatedCounty(the_request);
 
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
@@ -171,14 +171,14 @@ public class CVRExportImport extends AbstractCountyDashboardEndpoint {
         Main.LOGGER.info(imported + " CVRs parsed from file " + the_file.id() + 
                          " for county " + the_file.county().id());
         updateCountyDashboard(the_response, the_file, ImportStatus.SUCCESSFUL, imported);
-        the_file.setStatus(FileStatus.IMPORTED_AS_CVR_EXPORT);
-        Persistence.saveOrUpdate(the_file);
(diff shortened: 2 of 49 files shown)


nealmcb:
My testing shows that the code is a huge improvement - thanks!

server/eclipse-project/pom.xml
```diff
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>us.freeandfair.production</groupId>
 	<artifactId>colorado_rla</artifactId>
-	<version>1.0.3</version>
+	<version>1.0.4</version>
```
nealmcb: I vote to have this say something like 1.0.4.dev1 until it is merged into master. My test scripts now incorporate version numbers better into the metadata saved with each test, and this will clarify what was actually tested.
I'd actually prefer 1.0.4.dev-g1234567 with a commit id in there, but that would take some clever scripting work.
FWIW, these version number formats match the guidelines at https://docs.openstack.org/pbr/latest/user/semver.html  which is a take-off on semantic versioning 2.0, with Python PEP-440 compatibility.

dmzimmerman: I don't really want to do the clever scripting work, and having to commit a new commit after the PR is approved to change the version number from 1.0.4-dev to 1.0.4 is just asking for that step to be forgotten, repeatedly. But I'll see if I can think of something.

nealmcb: I figure we just need a checklist, and an extra minute, to accompany use of existing and/or improved scripts for doing a distribution (ala `script/build-release`). For now I'm also not volunteering to make a more clever script, so I'm just suggesting that e.g. for each release, at whatever time we decide "ship it", we 

* do a commit with the full release number 1.0.4
* build it
* tag it
* do a commit with the next release number 1.0.5.dev1.

Not much to forget then.
And when we do get around to doing a script, it can follow the PBR practice (documented along with the link up above) of actually just pulling the version from the tag itself, so no extra merges are necessary. That will be a small but helpful incentive to find or write one like that....

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
```diff
@@ -171,14 +171,14 @@ private void parseFile(final Response the_response, final UploadedFile the_file)
...
         updateCountyDashboard(the_response, the_file, ImportStatus.SUCCESSFUL, imported);
-        the_file.setStatus(FileStatus.IMPORTED_AS_CVR_EXPORT);
-        Persistence.saveOrUpdate(the_file);
         handleTies(the_response, the_file.county());
         final Map<String, Integer> response = new HashMap<String, Integer>();
         response.put("records_imported", imported);
         if (deleted > 0) {
           response.put("records_deleted", deleted);
         }
+        the_file.setStatus(FileStatus.IMPORTED_AS_CVR_EXPORT);
```
ranweiler: Why does moving this here make a difference wrt transactions? Were we e.g. sometimes running into database contention in `handleTies()`, and then when the calling code retried, the file update was spuriously appearing to be done already?

server/eclipse-project/pom.xml
ranweiler: Short version: I like @nealmcb's second suggestion, handling everything in one PR at the version boundary. I would only explicitly add that we needn't add a number to the `dev`-like suffix (e.g. just `1.0.5-dev` or `1.0.5-alpha` is fine). In any case, I don't think git hashes should be included in the version numbers.

ranweiler: (I see why @nealmcb might want `dev`/`alpha` suffix to be numbered, but IMO it would _ideal_ if that sort of tracking were managed at the level of the test scripts, which I'd guess would either know exactly what commit they are dealing with, or they are dealing with a release candidate on a test server, which I would consider a "test release", and thus its own natural tagging boundary).

server/eclipse-project/src/main/java/us/freeandfair/corla/endpoint/CVRExportImport.java
dmzimmerman: The short answer is yes, we had contention in `handleTies` (one of the only places where that happens), and so the file status would revert and not be reset. Now it gets committed with the overall transaction. Though that changes again with the next PR.

server/eclipse-project/pom.xml
dmzimmerman: I’ll add a “-dev” on it.

PR #854, opened 2017-10-26
Load tests

kiniry:


nealmcb:
.pyc files should not be included.

nealmcb:
It seems like we should not capture the reset time in the timing for each script.

kiniry:
External timing on the scripts presumes an empty DB.  I have been using some shell scripting to narrow log messages to get the precise timings you have seen in the load testing spreadsheet.  I will add a `someday` issue to improve this and make it fully automated.

test/load_tests/2-cdos.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+./main.py -u http://192.168.24.43/api/ reset dos_init > 0.log
```
dmzimmerman: Would it make sense to replace these scripts with scripts that take a command line parameter for the hostname?

nealmcb: A parameter would be better. Or my approach in autotest2.sh is to pass an environment variable, $URL.

PR #871, opened 2017-11-01
UX fixes, improvements

ranweiler:


client/src/notice.ts
```diff
@@ -1,28 +1,26 @@
...
+function warning(message: string, timeout: number = 0) {
     Toast.show({
         intent: Intent.WARNING,
         message,
         timeout,
     });
 }
 
-function ok(message: string, timeout: number = defaultNoticeTimeout) {
+function ok(message: string, timeout: number = 10000) {
```
dmzimmerman: Why is this now not a symbolic constant anymore?

ranweiler: Because there are now 3 degrees of freedom, instead 1 default. I'll go ahead and add two constants back to the config. It is nice to have it in one place.

PR #879, opened 2017-11-02
Fixes for Usernames and ASM Logging

### LINKED ISSUE #877: Add County Number to ASM Logging
ASM transitions currently do not log what county the ASM is for (if it's not for the DoS dashboard). They should.

### LINKED ISSUE #878: Usability Improvements for 2-Factor Auth
Apparently, two-factor auth is case-sensitive, but it is expected to be case-insensitive by county users. We should consider forcing user responses to upper case. Similarly, usernames are case-insensitive but we are treating them as case-sensitive; we should consider forcing them to lowercase.

dmzimmerman:
Closes #877 
Closes #878

### DIFF
--- a/server/eclipse-project/src/main/java/us/freeandfair/corla/asm/AbstractStateMachine.java
+++ b/server/eclipse-project/src/main/java/us/freeandfair/corla/asm/AbstractStateMachine.java
@@ -199,7 +199,9 @@ public abstract class AbstractStateMachine implements Serializable {
     // If we are in the right state then transition to the new state.
     if (the_transition.startStates().contains(my_current_state)) {
       my_current_state = the_transition.endState();
-      Main.LOGGER.info("ASM transition succeeded: " + the_transition); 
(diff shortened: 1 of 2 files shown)


sfsinger19103:
My notes from the 11/2 post-county-testing conference call says that we promised to force usernames to *upper* case, and @dmzimmerman 's comment in #878 also mentions *upper* case. Yet commit `2e88ae7` comment says force usernames to *lower* case. What gives?

nealmcb:
My recollection is that we wanted the grid challenge responses to be made upper case, and the usernames to be made lower case. And I see that matches what dmzimmerman said in #878.

As always, in case of confusion, this will clear everything up: [fourwordsalluppercase](https://www.youtube.com/watch?v=bLE7zsJk4AI)

dmzimmerman:
Actually, what we really wanted was to force usernames to lowercase, and to leave the grid challenge responses as-is because they are, as verified by us and CDOS multiple times, case insensitive.

PR #884, opened 2017-11-07
Commitment to AGPLv3 with classpath exception

sfsinger19103:
DO NOT MERGE THIS.  See https://github.com/FreeAndFair/ColoradoRLA/pull/899 instead.


LICENSE.md
```diff
@@ -1,22 +1 @@
...
-
-If you, the reader, have input on this decision, please add a comment
-to [Issue #452](https://github.com/FreeAndFair/ColoradoRLA/issues/452).
-
-Thus, as of now, this system is not licensed for any use whatsoever
-beyond its owner, the Colorado Department of State.
-
-
-
+The Colorado RLA Tool is Copyright (C) 2017 the Colorado Department of State, and is licensed under the [AGPLv3](https://www.gnu.org/licenses/agpl-3.0.en.html)  with a classpath exception.
```
nealmcb: I think we should follow normal practice and simply include the full AGPLv3. That keeps things self-contained, makes searches work better, etc.

sfsinger19103: OK, will do.

cjerdonek:
I don't have a great understanding of the classpath exception, but searching around online, it looks like if you want to license something under AGPLv3 with a classpath exception, it would best to include not just the AGPLv3 text, but also the text of the exception. I don't see that in the current PR when I search for the word "exception." Here is an example link with more info: https://opensource.stackexchange.com/questions/515/what-is-a-class-path-exception

cjerdonek:
Also, here seems to be the exception text from gnu.org:
https://www.gnu.org/software/classpath/license.html

And here is an example of its inclusion alongside another license (scroll to the very bottom):
http://openjdk.java.net/legal/gplv2+ce.html

LICENSE.md
```diff
@@ -1,22 +1,234 @@
...
+Copyright © 2007 Free Software Foundation, Inc. <https://fsf.org/> 
+Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed.
 
-Thus, as of now, this system is not licensed for any use whatsoever
-beyond its owner, the Colorado Department of State.
+Preamble
 
+The GNU Affero General Public License is a free, copyleft license for software and other kinds of works, specifically designed to ensure cooperation with the community in the case of network server software.
+
+The licenses for most software and other practical works are designed to take away your freedom to share and change the works. By contrast, our General Public Licenses are intended to guarantee your freedom to share and change all versions of a program--to make sure it remains free software for all its users.
```
cjerdonek: Nit: it would be nice if the lines were wrapped to a reasonable length, like 80 characters.

nealmcb:
Thank you, Chris!

I think the SPDX standard metadata for this license (assuming we do not allow later versions of AGPL) would be:

`SPDX-License-Identifier: AGPL-3.0 WITH Classpath-exception-2.0`

See 
* [GNU Affero General Public License v3.0 | Software Package Data Exchange (SPDX)](https://spdx.org/licenses/AGPL-3.0.html)
* [Classpath exception 2.0 | Software Package Data Exchange (SPDX)](https://spdx.org/licenses/Classpath-exception-2.0.html)

Any thoughts on specifying _or later version_ for this code?

nealmcb:
Since some parts of the code use a different license (e.g. rla_export is a standalone program which uses GPL-3.0), should we put multiple licenses at the top level, with names like "agpl-3.0.txt", and explain which applies to what in LICENSE.md?

It seems that some day we may want to use an "SPDX Document" to provide a machine-readable description like that:
[Using SPDX | Software Package Data Exchange (SPDX)](https://spdx.org/using-spdx)

nealmcb:
We'll want to update or look at these files which talk about licenses also:

- [ ] docs/05_license.md, 
- [ ]  client/package.json
- [ ]  docs/book.html
- [ ] server/eclipse-project/project_configuration/coloradorla_codetemplates.xml
- [ ] server/eclipse-project/script/rla_export/rla_export/__main__.py
- [ ] test/smoketest/main.py
- [ ] tools/delit.pl

cjerdonek:
> Since some parts of the code use a different license (e.g. rla_export is a standalone program which uses GPL-3.0), should we put multiple licenses at the top level

For repositories containing different files with different licenses, I would recommend making the files themselves the sources of truth. By this I mean keeping the statement of licenses with or close to the file(s) they apply to. For a single file, it can be at the top of the file. For multiple files, it would ideally be a LICENSE file in a directory containing only those files. Each within-project statement of license can be a canonical identifier combined with a brief single sentence or so (e.g. "This file is licensed under XXX. See the top-level LICENSE file for the full license text.") rather than the full text. The reason for this approach is that it will be much harder for things to get out of sync and not reconstructible in the face of file and directory renames, etc.

Under this model, the top-level LICENSE file could be a summary of the repository contents: At the top of the file, you could have a "table of contents" listing the name of each license and a rough description of which files it applies to. Below that table of contents, you could provide a concatenation of the full text of each license that appears in the project. This way all of the full license texts appear in a single file in the repository, and without duplication. Also, at the end of the table of contents at the top of the file, you could provide instructions that would let someone validate or reconstruct the table of contents, for example by listing the canonical strings one could search the code base for to confirm / find where each license is used.

cjerdonek:
By the way, here are step-by-step instructions from GNU itself for "what you need to do to release a program under one of our licenses": https://www.gnu.org/licenses/gpl-howto.html

Their instructions include things like adding a COPYING file with a copy of the license, and putting a (briefer) license notice near the top of each file.

sfsinger19103:
Hi Chris, thanks for the attention and suggestions. We're working on it.
Cheers, Stephanie

Stephanie Singer
Principled Project Lead
Free & Fair <http://freeandfair.us>
267-414-3119

cjerdonek:
Thanks for making progress on this, Stephanie.

Would it be possible for you to use a copy of the AGPL as it was originally formatted by GNU here?
https://www.gnu.org/licenses/agpl.txt

This way the license document will be the same as the original byte for byte, and it will be easier to tell that it is the same as the original. (It is also formatted a bit more nicely, with shorter lines, indenting, and centering, etc.) Lastly, this is also the link that GNU's step-by-step instructions I mentioned above points the reader to.

nealmcb:
Hmmm.  As the `rla_export` tool evolves toward one that actually generates, and could maintain in-real-time, the actual Audit Center, it might actually make sense to license it as AGPLv3 also....

test/smoketest/main.py
```diff
@@ -155,7 +155,7 @@
 
 
 __author__ = "Neal McBurnett <http://neal.mcburnett.org/>"
-__license__ = "GPLv3+"
+__license__ = "APLv3 with classpath exception"
```
nealmcb: Typo: AGPLv3....

cjerdonek:
FYI, it's hard to see what the changes are to comment on because 180+ commits were added that appear unrelated. Can that be cleaned up to allow comments on the most recent iteration?

nealmcb:
This PR has been superseded by https://github.com/FreeAndFair/ColoradoRLA/pull/899 which is ready for initial review and merge. More to be done there, but important to get the LICENSE.md file updated.

PR #896, opened 2017-11-12
Ballot Manifest Upload During CVR Import Fix

dmzimmerman:
There was a missing transition in the state machine: namely, the transition from `CVRS_IMPORTING` (and also `BALLOT_MANIFEST_OK_AND_CVRS_IMPORTING`) to `BALLOT_MANIFEST_OK_AND_CVRS_IMPORTING` when a ballot manifest is uploaded while a CVR is importing asynchronously. This PR adds that transition.

It also undoes the relaxation of the `comment` field name on ACVR uploads that I had put in because I had thought the client bug related to comments was related to field naming rather than just not including the comments in the submission.


dmzimmerman:
Closing this, because it's clearly not going to be merged at this point.

PR #906, opened 2018-01-08
Remove license line from almost all files

sfsinger19103:


nealmcb:
Hmmm - don't we want to mention the license in the files somewhere? Just take out the "Classpath" part?
Let me review what Chris advised.

nealmcb:
See e.g. https://github.com/FreeAndFair/ColoradoRLA/pull/884#issuecomment-346979966
where Chris notes that FSF advises putting a (briefer) license notice near the top of each file.

https://www.gnu.org/licenses/gpl-howto.html
> Put a license notice in each file.

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
```diff
@@ -4,7 +4,6 @@
  * @title ColoradoRLA
  * @created Jul 19, 2017
  * @copyright 2017 Colorado Department of State
- * @license GNU Affero General Public License v3 with Classpath Exception
  * @creator Daniel M. Zimmerman <dmz@freeandfair.us>
```
nealmcb: I think that as Chris suggests we should follow FSF advice:

https://www.gnu.org/licenses/gpl-howto.html
> Put a license notice in each file.

I'd substitute this instead:
> @license SPDX-License-Identifier: AGPL-3.0-only
or this
> @license SPDX-License-Identifier: AGPL-3.0-or-later

nealmcb:
If a new version of AGPL comes out, should we be able to use it? I like to trust FSF and guess a fix might be important some day.  But suppose some horrid takeover could take place....
Given the large project and sometimes confusing licensing, SPDX seems useful to me.

I.e., should we say 
> @license SPDX-License-Identifier: AGPL-3.0-only
or this
> @license SPDX-License-Identifier: AGPL-3.0-or-later

sfsinger19103:
I don't know.

kiniry:
I'm just now seeing this discussion while I was working on something else and then rebasing this.

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
kiniry: I agree.  I will make this update now.

kiniry:
@ranweiler We need a license declaration in every file you wrote in the front-end too.

ranweiler:
@kiniry ack. I can add that this weekend.

server/eclipse-project/src/main/java/us/freeandfair/corla/Main.java
kiniry: You are welcome to review again, @nealmcb.  The push is making its way through CI.

nealmcb:
I think we can either include the client license line changes or not with this PR. Other changes still needed are listed in #899

PR #918, opened 2018-09-10
genelect.py: Generate test files for RLAtool

nealmcb:
See file block comment (or pydoc3 genelect) for documentation,
and genelect.py -h for help


nealmcb:
Sorry.  Sigh - github makes it really easy to enter a pull request on the wrong repo....
But this is a handy little genelect.py script that would be easy to pull all by itself.

